PR #6, opened 2026-02-04
RHF Form Components

pohanson:
- **Install RHF and React-Icon**
- **Add `RHFInput` which wraps the various form input**
- **Add `RHFTextArea` for form textarea**
- **Add `RHFRadio` component for radio input**
- **Add `Button` component**

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -1,8 +1,4 @@
-import {
-  ButtonSize,
-  ButtonVariant,
-  getButonTwClassNames,
-} from "./buttonStyles";
+import { ButtonSize, ButtonVariant, getButonTwClassName } from "./buttonStyles";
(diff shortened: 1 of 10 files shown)


src/components/Button.tsx
```diff
@@ -0,0 +1,54 @@
...
+  variant?: ButtonVariant;
+  size?: ButtonSize;
+  loading?: boolean;
+}) {
+  return (
+    <button
+      onClick={onClick}
+      type={type}
+      title={title}
+      disabled={loading}
```
nathantew14: can we have a differentiation between loading and disabled? maybe some loading indicator

nathantew14 (changes requested):

nathantew14 (approved):
Lgtm from just looking, will have to refine further when we actually start using the components

PR #15, opened 2026-02-16
Move auth logic to proxy

noahseethorcodes:
This PR moves our auth model from primarily client-side protection to a server-side, proxy-based guard for the entire app (with `/login` as the only public route).

Previously, we relied on a `withAuth` HOC and `useAuth` hook to protect pages in the browser. Each protected page rendered on the client, then ran Supabase `getSession()` and `onAuthStateChange` to decide whether to redirect to `/login`. This worked functionally, but it meant that:

- The server would still send HTML for protected routes before we knew if the user was authenticated.
- Route protection logic was split across client and server, which made reasoning about access a bit harder.
- We were doing extra auth work on the client that we could handle once centrally.

With this PR, we’ve upgraded `proxy.ts` so it now both refreshes the Supabase session **and** enforces route-level access at the edge. On every request, the proxy:

- Creates a Supabase server client from the incoming cookies.
- Calls `supabase.auth.getUser()` to transparently refresh tokens and update cookies.
- Treats all routes as protected by default, except `/login`.
- Redirects unauthenticated users hitting any non-`/login` path to `/login`, preserving the intended destination in a `redirectTo` query param.
- Redirects authenticated users away from `/login` to `/dashboard`, so they don’t see the login screen when already signed in.

The main benefit is that our access control is now enforced before any protected page or data is rendered, which is a better fit for a patient-facing system. We have a single, central place (the proxy) that defines what “authenticated” means and which routes are reachable in each state. The client-side auth utilities can be simplified to focus on UX (showing the current user, reacting to sign-outs, etc.) rather than acting as the primary security boundary.

### DIFF
--- a/src/lib/session.tsx
+++ b/src/lib/session.tsx
@@ -33,8 +33,7 @@ import { Session } from "@supabase/supabase-js";
  * @see {@link https://supabase.com/docs/guides/auth} for Supabase auth details.
  * @since 1.0.0
  */
-export function useAuth() {
-  const router = useRouter();
+export function useSession() {
   const [session, setSession] = useState<Session | null>(null);
   const [loading, setLoading] = useState(true);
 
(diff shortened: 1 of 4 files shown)


pohanson (approved):
LGTM

pohanson (approved):

PR #51, opened 2026-02-21
#48: Add patient image url to patient table

### LINKED ISSUE #48: Add Cloudinary for Patient Picture

noahseethorcodes:
Progresses issue #48 

This PR adds the `patient_image_url` to the `patients` table. It also adds
- A url to a mock photo for the seeding data (note all 15 rows have the same url to the same mock image)
- Logic on the patient router to process the URL
- A frontend component to display the image

Patient page now:
Screenshot 2026-02-21 at 5 44 05 PM


src/server/routers/patients_router.ts
```diff
@@ -41,6 +52,7 @@ export const patientsRouter = router({
           bs2: patients.bs2,
           drugAllergy: patients.drugAllergy,
           sabaiCard: patients.sabaiCard,
+          patientImageUrl: patients.patientImageUrl,
```
pohanson: Would the `patientImageUrl` include the `cloudinaryUrl`?

pohanson (changes requested):

src/server/routers/patients_router.ts
noahseethorcodes: oops forgot to combine with the cloudinary url prefix to get the full url. have pushed a commit to fix this. thanks for the spot!

src/db/schema.ts
```diff
@@ -53,6 +53,7 @@ Patients Table:
 - poor: Indicates if the patient has a POOR card or not.
 - bs2: Indicates if the patient has a BS2 card or not.
 - sabaiCard: Indicates if the patient has a Sabai card or not.
+- patientImageUrl: Cloudinary URL of the patient's image.
```
nathantew14: if it's not gonna be the full url can we avoid naming the column xxUrl? 
> patientImageId or patientImagePublicId — the latter is actually what Cloudinary calls it in their own terminology, so it's quite descriptive.

^ from Claude

nathantew14 (changes requested):
Just one small nitpick on the naming
I will deal with the env var in my own PR

src/server/routers/patients_router.ts
```diff
@@ -19,10 +24,16 @@ export const patientsRouter = router({
...
+        patientImageUrl: patients.patientImageUrl,
       })
       .from(patients);
 
-    return result;
+    return result.map((patient) => ({
+      ...patient,
+      patientImageUrl: patient.patientImageUrl
+        ? `${cloudinaryUrl}/${patient.patientImageUrl}`
+        : null,
```
nathantew14: Maybe extract this out to a function?

src/server/routers/patients_router.ts
```diff
@@ -41,11 +52,16 @@ export const patientsRouter = router({
...
           sabaiCard: patients.sabaiCard,
+          patientImageUrl: patients.patientImageUrl,
         })
         .from(patients)
         .where(eq(patients.id, input.id))
         .limit(1);
 
+      if (result) {
+        result.patientImageUrl = `${cloudinaryUrl}/${result.patientImageUrl}`;
+      }
```
nathantew14, line 73: this one doesn't have the null check unlike the previous one

src/db/schema.ts
nathantew14: any reason why we're using a yesNoEnum rather than a boolean?

src/db/schema.ts
noahseethorcodes: Implemented suggested change. Note in PR #52 the envar name changes to `CLOUDINARY_URL_PREFIX` as `CLOUDINARY_URL` is reserved for the cloudinary sdk envar

src/db/schema.ts
noahseethorcodes: main reason at point of implementation is that i followed the legacy data. i agree boolean is cleaner and better and will make the change

nathantew14 (changes requested):
looks good!

nathantew14 (approved):
submitted previous request for changes by accident

pohanson (approved):

PR #55, opened 2026-02-25
Medication-related schemas and routers

### LINKED ISSUE #34: Medication Schema
Plan out the schema here, and check in with the team

**Schema**:

keiran0:
Closes #34 

Why:

In the old repo, the medication types and stock were in the same table, and medicine side used an Excel sheet to track locations and expiry date of each medication stock. To avoid manual entry and correction, they requested for the stocktaking and tracking of these details to be handled by the app. There were also medicines that were interchangeable (active ingredient and delivery method) with each other but not properly represented. The separation of the tables in the new schema will help with the new feature requested and prepares for the new pharm pages.

Summary:

- Adds new medication related schemas
- Adds migrations for the database as well as seeds for the initial data.
- ~~Dummy frontend was also included to view the data present in the database~~ Removed as the main placeholder frontend was also replaced. 
- Add routers for medication (brand, active ingredient, stock)


keiran0:
Oops. Hold off on merging until I correct the code to use zod-form-data.

src/db/schema.ts
```diff
@@ -73,3 +81,52 @@ export const patients = pgTable("patients", {
 
 export type Patient = typeof patients.$inferSelect;
 export type NewPatient = typeof patients.$inferInsert;
+
```
derrikzzz, line 83: Perhaps would it be better to have table documentation comments on the top to describe the schema of the table? Something like
```typescript
/*
Medication Active Ingredients Table:
- id: Primary key, auto-incrementing integer
- name: Name of the active ingredient (e.g., "Paracetamol")
- unitOfMeasurement: Unit used for measuring this ingredient (e.g., "mg", "ml")
- fallBelow: Threshold quantity that triggers low stock alerts
*/
```
 
Can refer to patients table above as an example!

src/db/schema.ts
```diff
@@ -73,3 +81,52 @@ export const patients = pgTable("patients", {
...
+  name: varchar("name", { length: 255 }).notNull(),
+  activeIngredientId: integer("active_ingredient_id")
+    .notNull()
+    .references(() => medicationActiveIngredients.id, {
+      onDelete: "restrict",
+    }),
+});
+
+export type MedicationBrands = typeof medicationBrands.$inferSelect;
+export type NewMedicationBrands = typeof medicationBrands.$inferInsert;
```
derrikzzz: Should we standardize the inferred types to be singular? like 
```
export type MedicationBrand = typeof medicationBrands.$inferSelect;
export type NewMedicationBrand = typeof medicationBrands.$inferInsert;
```

src/db/schema.ts
```diff
@@ -6,10 +6,18 @@ import {
   timestamp,
   text,
   pgEnum,
+  integer,
 } from "drizzle-orm/pg-core";
 
 // Define enums
 export const genderEnum = pgEnum("gender", ["male", "female"]);
+export const yesNoEnum = pgEnum("yes_no", ["yes", "no"]);
```
derrikzzz: I don't think that this `yesNoEnum` is used in this file, if we plan to use it in the future to replace some boolean fields then perhaps adding a comment here might be better?

src/db/schema.ts
```diff
@@ -73,3 +81,52 @@ export const patients = pgTable("patients", {
...
+  id: serial("id").primaryKey(),
+  medicationBrandId: integer("medication_brand_id")
+    .notNull()
+    .references(() => medicationBrands.id, {
+      onDelete: "restrict",
+    }),
+  quantity: integer("quantity").notNull().default(0),
+  expiry: timestamp("expiry"),
+  location: varchar("location", { length: 255 }),
+  state: medicationStatusEnum("state").default("active"),
```
derrikzzz: Should `state` be renamed to something like `stockStatus` for clearer variable name?

derrikzzz (changes requested):
Solid work on the medication management feature! Just a few minor notes on code consistency and documentation:

1. Table Documentation: Adding of descriptive comments to the new medication tables
2. Type Naming Consistency for the inferred types
3. Unused Enum: The `yesNoEnum` is defined but not used, adding a comment would be helpful if we plan to use it in the future to replace some boolean fields

src/db/schema.ts
keiran0: Thanks, added

src/db/schema.ts
keiran0: Good catch, resolved

src/db/schema.ts
keiran0: I believe it might be an old implementation from somewhere, but I will remove it in this PR too for convenience

src/db/schema.ts
keiran0: You're right, that sounds cleaner, resolveed

supabase/migrations/0004_third_angel.sql
```diff
@@ -0,0 +1,25 @@
...
+	"active_ingredient_id" integer NOT NULL
+);
+--> statement-breakpoint
+CREATE TABLE "medication_stock" (
+	"id" serial PRIMARY KEY NOT NULL,
+	"medication_brand_id" integer NOT NULL,
+	"quantity" integer DEFAULT 0 NOT NULL,
+	"expiry" timestamp,
+	"location" varchar(255),
+	"state" "medication_status" DEFAULT 'active'
```
derrikzzz: Perhaps should we have a migration script to rename `state` to `stock_status`, for consistency!

derrikzzz (changes requested):
Good job with the changes! Just some feedback on the necessity of generating a migration to rename the column `state` to `stock_status` in the `medication_stock` table!

supabase/migrations/0004_third_angel.sql
keiran0: I deleted that migration and recreated it to avoid more conflicts with 003, since it's been really long since I last created that migration, the column name has been changed as well. Thanks

derrikzzz (approved):
Good job with the changes! Looks good to merge! 👍

PR #57, opened 2026-04-22
Add visits schema and routers

derrikzzz:
## Summary of changes
- swap positions of label text and checkbox for the card fields
- Add visits schema
- Add visitsRouter with `create`, `list`, and `listByPatient` procedures
- Add `createWithVisit` mutation to patientsRouter that atomically creates a patient and their initial visit in a transaction, with Cloudinary cleanup on failure
- Update `scanFace.tsx` to use `createWithVisit` mutation and include village code selection


keiran0 (changes requested):
The merge with dev might need to be redone as the `visits` table is missing after migration, going forwards we should prioritise merging changes to schema oops.

keiran0 (approved):
LGTM!

src/server/routers/visits_router.ts
```diff
@@ -0,0 +1,119 @@
...
+      return result;
+    }),
+
+  // Create new visit
+  create: protectedProcedure
+    .input(
+      z.object({
+        patientId: z.number().int(),
+        villageCodeId: z.number().int(),
+        date: z.date().optional(),
```
keiran0: I believe this might be correct because superjson might handle this, but the implementation also conflicts with `patientsRouter`, let's KIV and come back to this (and all dates) after we implement some forms.

keiran0 (changes requested):
Adding a file to the seeds would be nice, otherwise LGTM!

keiran0 (changes requested):
I believe the changes to the the `scripts` folder, `package.json` and `README.md` isn't necessary, just add in a sql file into the `supabase/seeds` folder.

`config.toml` line 60-65 will automatically seed the database using all the files inside this directory when the database is reset.

keiran0 (approved):
LGTM!

PR #61, opened 2026-05-11
Fix/cloudinary url

derrikzzz:
## Problem

- The `uploadToCloudinary` function was returning URLs with incorrect format that included duplicated `image/upload` method, which could have caused issues when constructing the URLs

## Solution
- Fixed return value to only return the path portion without the duplicated `image/upload/` prefix

## Breaking Changes
- None, this is a bug fix

## Tests
- Fill up a dummy patient and capture photo, then head over to patients page to see if image gets rendered

## Future TODOs
- N.A.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -6,5 +6,7 @@ SUPABASE_SECRET_KEY=your-secret-key-here
 DATABASE_URL=postgresql://postgres:postgres@127.0.0.1:54322/postgres
 
 # CLOUDINARY
-CLOUDINARY_URL_PREFIX=your_cloudinary_url_prefix # prefix for image links
-CLOUDINARY_URL=cloudinary://API_KEY:API_SECRET@CLOUD_NAME # for cloudinary API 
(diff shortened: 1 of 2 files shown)


naa-siuuuu-ff (commented):
LGTM!

keiran0:
Hold off on merging first, let's check something, can you remove `CLOUDINARY_URL_PREFIX`'s duplicated 'image/upload' at the end of the variable and retry again on dev branch?

keiran0:
I believe it could have been an error with the `.env` file instead that was causing the duplicats

naa-siuuuu-ff (approved):

keiran0:
In the `.env.example`, I believe the last '/' in `CLOUDINARY_URL_PREFIX` needs to be omitted. After following the example, if the webpage is inspected, the `src` attribute of the `img` tag has a duplicate '/' in it. But visually, there's no issue.

keiran0 (approved):
LGTM!

PR #62, opened 2026-05-13
Fix column name for medication stock seed

keiran0:
This PR:
- fixes the medication stock seed file as the column name wasn't changed after changing from `status` to `stock_status`.
- adds in the proper column name that was somehow lost when migration 4 was created (stashed somewhere possibly?) and thus not reflected correctly in the current code (current code in dev wrong, migration in dev correct, but now this code matches the code that was used to generate the correct migration)
- renames the column throughout the router as well for all methods

### DIFF
--- a/src/server/routers/medication_stock_router.ts
+++ b/src/server/routers/medication_stock_router.ts
@@ -14,7 +14,7 @@ export const medicationStockRouter = router({
         quantity: medicationStock.quantity,
         expiry: medicationStock.expiry,
         location: medicationStock.location,
-        state: medicationStock.stockStatus,
+        stock_status: medicationStock.stockStatus,
       })
       .from(medicationStock);
     return result;
@@ -27,7 +27,7 @@ export const medicationStockRouter = router({
         quantity: zfd.numeric(z.number().int()),
         expiry: zfd.text(z.coerce.date()),
         location: zfd.text(),
-        state: zfd.text(z.enum(medicationStatusEnum.enumValues)),
+        stock_status: zfd.text(z.enum(medicationStatusEnum.enumValues)),
       }),
     )
     .mutation(async ({ input }) => {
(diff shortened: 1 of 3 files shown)


src/server/routers/medication_stock_router.ts
```diff
@@ -27,7 +27,7 @@ export const medicationStockRouter = router({
         quantity: zfd.numeric(z.number().int()),
         expiry: zfd.text(z.coerce.date()),
         location: zfd.text(),
-        state: zfd.text(z.enum(medicationStatusEnum.enumValues)),
+        stock_status: zfd.text(z.enum(medicationStatusEnum.enumValues)),
```
derrikzzz: should we use camelCase for this! `stockStatus` instead of `stock_status`, since this prob follows conventions (can refer to the distinction between `medicationBrandId` and `medication_brand_id` in schema

src/server/routers/medication_stock_router.ts
```diff
@@ -62,7 +62,7 @@ export const medicationStockRouter = router({
         quantity: zfd.numeric(z.number().int().optional()),
         expiry: zfd.text(z.coerce.date().optional()),
         location: zfd.text(z.string().optional()),
-        state: zfd.text(z.enum(medicationStatusEnum.enumValues).optional()),
+        stock_status: zfd.text(z.enum(medicationStatusEnum.enumValues).optional()),
```
derrikzzz: Has been addressed by above comment!

derrikzzz (changes requested):
Good job on noticing the fixing of the column name! Just some minor nit on the necessity on changing to `camelCase` for `stock_status` in the router file!

src/server/routers/medication_stock_router.ts
keiran0: Thanks so much for catching this, too much copying and pasting was done here hahaha

derrikzzz (approved):
No nits about this, looks good to merge!

PR #66, opened 2026-05-18
Medication dummy frontend

keiran0:
This PR creates 3 new pages:
1) Medication stock
2) Medication active ingredients
3) Medication brands

2 and 3 can only be accessed from the medication stock page, as they are not expected to be handled by people outside of PL. There is a button to access these 2 pages from the medication stock page.

These pages only display the content present in the tables, without joining. It is a placeholder for the actual pages, which will be implemented later. 

Minor change: Added `.editorconfig` to stop prettier and linter from screaming at Windows users.


src/pages/medication-active-ingredients/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+              strokeWidth="4"
+            ></circle>
+            <path
+              className="opacity-75"
+              fill="currentColor"
+              d="M4 12a8 8 0 018-8V0C5.373 0 0 5.373 0 12h4zm2 5.291A7.962 7.962 0 014 12H0c0 3.042 1.135 5.824 3 7.938l3-2.647z"
+            ></path>
+          </svg>
+          Loading active ingredients...
+        </div>
```
derrikzzz: Perhaps this can be abstracted as a component by itself seeing that a similar implementation is found in `src/pages/medication-brands/index.tsx` and `src/pages/medication-stock/index.tsx` too?

src/pages/medication-active-ingredients/index.tsx
```diff
@@ -0,0 +1,121 @@
+import { trpc } from "@/utils/trpc";
+import Breadcrumbs from "@/components/Breadcrumbs";
+
+function MedicationActiveIngredientsBasePage() {
+  const { data: ingredients, isLoading } =
```
derrikzzz, line 6: Right now i don't think we have any error handling in the codebase, would it make sense to introduce here!

src/pages/medication-stock/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+                <tr
+                  key={item.id}
+                  className="hover:bg-slate-50 transition-colors"
+                >
+                  <td className="px-6 py-4 whitespace-nowrap">
+                    <div className="flex flex-col items-center gap-2">
+                      <span className="text-sm font-medium text-slate-900">
+                        {item.id}
+                      </span>
+                    </div>
```
derrikzzz: UI concern, should it be `items-left` instead of `items-center`? (applies to the tables)

derrikzzz (changes requested):
Looks good mostly, just some feedback on the code duplication and potentially on whether we want to introduce error handling for fetching of data!

src/pages/medication-active-ingredients/index.tsx
keiran0: I removed it and created a new LoadingSpinner component. Since I got this from the patients page, so I also replaced that with the new component, thanks!

src/pages/medication-stock/index.tsx
keiran0: I suppose it does look a lot better when aligned, fixed

src/pages/medication-active-ingredients/index.tsx
keiran0: As discussed, created an issue #67 ~~haha~~ and will settle in another PR.

src/pages/medication-active-ingredients/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+                    </div>
+                  </td>
+                </tr>
+              ))}
+            </tbody>
+          </table>
+        </div>
+      ) : (
+        <div className="p-12 text-center text-slate-500">
+          No stock found. Seed the database or add a new record.
```
derrikzzz: This string "No stock found. Seed the database or add a new record" is being used in multiple files, would it make sense to be extracted as a constant, or should we make each message specific to each page type?

derrikzzz (commented):
Looks good mostly! Just a minor nit on the duplicated message in the `index.tsx` files!

src/pages/medication-active-ingredients/index.tsx
keiran0: Fixed

.editorconfig
```diff
@@ -0,0 +1,4 @@
+root = true
```
derrikzzz, line 1: Kudos on using this to enforce LF line endings across all files!

derrikzzz (approved):
Looks good to merge, thanks for resolving the config on lf vs crlf!

PR #78, opened 2026-05-27
Frontend refactoring and error handling

keiran0:
This PR:

- Refactors the frontend code to use early return for (previously) conditional rendering
- Makes use of the new Table components for all the pages that use it. Note that this causes a minor change in UI for the village codes page
- Formats old components (`pnpm run format`) that have not been formatted yet
- Adds `isError` to any `index.tsx` that makes use of `useQuery`. This does not apply to the `village-codes` page as the entire thing will be refactored later.
- Adds an `isError` check to all components that have been modified by the previous point, and have it display a general error message. The actual logging of errors will be handled in the future.

### DIFF
--- a/src/pages/medication-active-ingredients/index.tsx
+++ b/src/pages/medication-active-ingredients/index.tsx
@@ -65,7 +65,6 @@ function MedicationActiveIngredientsBasePage() {
               "Unit of Measurement",
               "Fall Below",
             ]}
-            className="bg-slate-50"
           />
           <tbody className="bg-white divide-y divide-slate-200">
             {ingredients.map((ingredient) => (
--- a/src/pages/medication-active-ingredients/index.tsx
+++ b/src/pages/medication-active-ingredients/index.tsx
@@ -58,7 +58,15 @@ function MedicationActiveIngredientsBasePage() {
       <Header />
       <div className="overflow-x-auto">
         <table className="min-w-full divide-y divide-slate-200">
-          <TableHeader headers={["Active Ingredient ID", "Active Ingredient Name", "Unit of Measurement", "Fall Below"]} className="bg-slate-50" />
+          <TableHeader
+            headers={[
+              "Active Ingredient ID",
+              "Active Ingredient Name",
+              "Unit of Measurement",
+              "Fall Below",
+            ]}
+            className="bg-slate-50"
+          />
           <tbody className="bg-white divide-y divide-slate-200">
             {ingredients.map((ingredient) => (
               <TableRow key={ingredient.id}>
--- a/src/pages/medication-active-ingredients/index.tsx
+++ b/src/pages/medication-active-ingredients/index.tsx
@@ -36,38 +36,24 @@ function MedicationActiveIngredientsBasePage() {
     isError,
   } = trpc.medicationActiveIngredientsRouter.list.useQuery();
 
-  if (isError) {
-    return (
-      <div className="min-h-screen flex-1 p-8">
-        <Header />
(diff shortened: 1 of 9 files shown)


src/pages/medication-active-ingredients/index.tsx
```diff
@@ -1,98 +1,112 @@
...
+              "Active Ingredient Name",
+              "Unit of Measurement",
+              "Fall Below",
+            ]}
+          />
+          <tbody className="bg-white divide-y divide-slate-200">
+            {ingredients.map((ingredient) => (
+              <TableRow key={ingredient.id}>
+                <TableCell>
+                  <div className="flex flex-col items-left gap-2">
```
derrikzzz: Is `items-left` valid tailwind here? https://tailwindcss.com/docs/align-items

src/pages/medication-stock/index.tsx
```diff
@@ -1,127 +1,140 @@
...
+            Medication Stock
+          </h1>
+          <p className="mt-2 text-slate-600">Manage medication stock.</p>
+          <button
+            type="submit"
+            className={`flex-1 px-4 py-2 rounded-lg font-medium bg-red-300`}
+          >
+            <Link
+              href="/medication-brands"
+              className="group flex items-center gap-2 p-2 rounded-md"
```
derrikzzz: Is the tailwind class `gap-2` necessary here? There's only 1 child element here

src/pages/medication-active-ingredients/index.tsx
keiran0: Nope, guess I was trying to put `items-start` but even so that would be invalid since it has only one child. Removed in the latest commit

src/pages/medication-stock/index.tsx
keiran0: Fixed, too much copy and pasting from sidebar (where the usage of `gap` actually makes sense), good catch

src/pages/medication-stock/index.tsx
```diff
@@ -1,127 +1,140 @@
...
-              <Link
-                href="/medication-active-ingredients"
-                className="group flex items-center gap-2 p-2 rounded-md"
-              >
-                <span>Manage Active Ingredients</span>
-              </Link>
-            </button>
-          </div>
+              <span>Manage Active Ingredients</span>
+            </Link>
```
derrikzzz, line 44: Minor nit: will it be better to increase spacing between the 2 buttons here?

keiran0: That will be a future fix, the buttons will be on the right side of the heading area, and then they will be set with appropriate colours. That's a future commit when I finally add in a third button to add new stock and a new column for editing each row of the stock. This PR will be focusing more on the refactor and code structure rather than the implementation of specific pages

src/pages/medication-brands/index.tsx
```diff
@@ -1,86 +1,90 @@
...
+  function renderContent() {
+    if (isError) {
+      return <h1 className="text-red-500">An error has occurred!</h1>;
+    }
+
+    if (isLoading) {
+      return <LoadingSpinner message="Loading brands..." />;
+    }
+
+    if (!brands || brands.length == 0) {
```
derrikzzz: Minor nit: should we use strict equality for this?

keiran0: Fixed, thanks for the catch!

derrikzzz (approved):
Looks good to merge, thanks for the refactoring help here!

PR #81, opened 2026-05-27
Modal custom component

keiran0:
This PR adds a modal component and refactors the village codes page to use it. Closes #80

### DIFF
--- a/src/components/interactive/Modal.tsx
+++ b/src/components/interactive/Modal.tsx
@@ -24,14 +24,12 @@ export default function Modal({ children, isOpen, setIsOpen }: ModalProps) {
       return;
     }
 
-    if (isOpen) {
-      if (!dialog.open) {
-        dialog.showModal();
-      }
-    } else {
-      if (dialog.open) {
-        dialog.close();
-      }
+    if (isOpen && !dialog.open) {
+      dialog.showModal();
+    }
+
+    if (!isOpen && dialog.open) {
+      dialog.close();
     }
   }, [isOpen]);
 
--- a/src/components/interactive/Modal.tsx
+++ b/src/components/interactive/Modal.tsx
@@ -1,0 +1,45 @@
+"use client";
+
+import { createPortal } from "react-dom";
+import { FaWindowClose } from "react-icons/fa";
+import { ReactNode } from "react";
+
+export interface ModalProps {
+  children: ReactNode;
+  isOpen: boolean;
+  setIsOpen: React.Dispatch<React.SetStateAction<boolean>>;
+}
+
+/**
+ * A modal component. This component mounts into `#modal-root`.
+ * @param {ReactNode} children - The contents of the modal
+ * @param {boolean} isOpen - Whether the modal is open or closed
+ * @param {React.Dispatch<React.SetStateAction<boolean>>} setIsOpen - The set state function provided by React's useState hook
+ */
+export default function Modal({ children, isOpen, setIsOpen }: ModalProps) {
+  if (!isOpen) {
+    return null;
+  }
+
+  return createPortal(
+    <div
+      className="fixed inset-0 bg-black/50 flex items-center justify-center p-4 z-50"
+      onClick={() => setIsOpen(false)}
+    >
+      <dialog
+        open={isOpen}
+        className="bg-white rounded-xl shadow-xl p-6 w-full max-w-md relative flex min-h-60 flex-col justify-between"
+        onClick={(e) => e.stopPropagation()}
+      >
+        <button
+          onClick={() => setIsOpen(false)}
+          className="absolute top-4 right-4"
+        >
+          <FaWindowClose className="text-red-500" />
+        </button>
+        {children}
+      </dialog>
+    </div>,
+    document.getElementById("modal-root")!,
+  );
+}
--- a/src/components/interactive/Modal.tsx
+++ b/src/components/interactive/Modal.tsx
@@ -5,58 +5,44 @@ import { ReactNode, useEffect, useRef } from "react";
 
 export interface ModalProps {
   children: ReactNode;
-  isOpen: boolean;
-  setIsOpen: React.Dispatch<React.SetStateAction<boolean>>;
+  onClose: () => void;
 }
 
 /**
  * A modal component. This component mounts into `#modal-root`.
  * @param {ReactNode} children - The contents of the modal
- * @param {boolean} isOpen - Whether the modal is open or closed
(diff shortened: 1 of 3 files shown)


src/components/interactive/Modal.tsx
```diff
@@ -0,0 +1,45 @@
...
+          className="absolute top-4 right-4"
+        >
+          <FaWindowClose className="text-red-500" />
+        </button>
+        {children}
+      </dialog>
+    </div>,
+    document.getElementById("modal-root")!,
+  );
+}
```
derrikzzz, line 50: Minor nit: should we have a useEffect here to handle the closing of the modal using escape key?

src/components/interactive/Modal.tsx
```diff
@@ -0,0 +1,45 @@
...
+  if (!isOpen) {
+    return null;
+  }
+
+  return createPortal(
+    <div
+      className="fixed inset-0 bg-black/50 flex items-center justify-center p-4 z-50"
+      onClick={() => setIsOpen(false)}
+    >
+      <dialog
```
derrikzzz: Is there a reason on why we are not using `showModal()` method here? (https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/showModal)

src/pages/_app.tsx
```diff
@@ -36,10 +36,13 @@ function App({ Component, pageProps }: AppPropsWithLayout) {
...
-    <Component
-      {...pageProps}
-      key={isPatientRoute ? router.asPath : undefined}
-    />,
+    <>
+      <Component
+        {...pageProps}
+        key={isPatientRoute ? router.asPath : undefined}
+      />
+      <div id="modal-root" className="h-screen w-screen fixed" />
```
derrikzzz: Would the `h-screen w-screen` be necessary here?

src/components/interactive/Modal.tsx
```diff
@@ -0,0 +1,45 @@
...
+        open={isOpen}
+        className="bg-white rounded-xl shadow-xl p-6 w-full max-w-md relative flex min-h-60 flex-col justify-between"
+        onClick={(e) => e.stopPropagation()}
+      >
+        <button
+          onClick={() => setIsOpen(false)}
+          className="absolute top-4 right-4"
+        >
+          <FaWindowClose className="text-red-500" />
```
derrikzzz: Minor nit: could we make the button bigger and have hover effect for interactivity?(https://tailwindcss.com/docs/transition-property)

src/components/interactive/Modal.tsx
keiran0: Fixed in current implementation, dialog elements have an `onClose` that will account for that. Thanks for the catch

src/components/interactive/Modal.tsx
keiran0: Fixed

src/pages/_app.tsx
keiran0: `#modal-root` is no longer required due to the usage of the dialog and removal of `createPortal` in the latest commit

src/components/interactive/Modal.tsx
keiran0: Fixed!

src/components/interactive/Modal.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+    if (isOpen) {
+      if (!dialog.open) {
+        dialog.showModal();
+      }
+    } else {
+      if (dialog.open) {
+        dialog.close();
+      }
+    }
```
derrikzzz: Minor nit: would there be a way to keep the happy path simple here instead of nested if statements?

src/pages/settings/village-codes.tsx
```diff
@@ -113,99 +114,105 @@ function VillageCodesPage() {
...
-                )}
+          <Modal
+            isOpen={isEditing}
+            setIsOpen={(open) => {
+              if (!open) {
+                closeForm();
+              } else {
+                setIsEditing(true);
+              }
+            }}
```
derrikzzz: For readability, would it make sense if we take the following approach:
```
isOpen={isEditing}
onClose={closeForm}
```
With this approach instead of an if check, it requires the Modal component to take in onClose prop where `onClose: () => void;`

src/components/interactive/Modal.tsx
keiran0: Fixed

keiran0:
I have reworked the props, there is now an `onClose` prop and `isOpen` is now redundant (we need to conditionally render it in the page using the component anyway)

The backdrop triggered by `showModal` will be done with `requestAnimationFrames`

src/pages/settings/village-codes.tsx
keiran0: Added the prop, thanks!

src/components/interactive/Modal.tsx
```diff
@@ -0,0 +1,48 @@
...
+ * @param {() => void} onClose - The function that runs when the Modal is closed or cancelled
+ */
+export default function Modal({ children, onClose }: ModalProps) {
+  const dialogRef = useRef<HTMLDialogElement>(null);
+
+  useEffect(() => {
+    const dialog = dialogRef.current;
+    requestAnimationFrame(() => {
+      dialog!.showModal();
+    });
```
derrikzzz: Minor nit: for readability, would changing it to below achieve the same effect?
```
if (dialog) {
    requestAnimationFrame(() => {
      dialog.showModal();
    });
  }
```

src/pages/settings/village-codes.tsx
```diff
@@ -113,99 +114,104 @@ function VillageCodesPage() {
...
-                      placeholder="e.g. V001"
-                    />
-                  </div>
-                )}
+          <Modal
+            onClose={() => {
+              closeForm();
+              setIsEditing(false);
+            }}
+          >
```
derrikzzz: Will it be possible to have 
```
 <Modal onClose={closeForm}>
```
since closeForm technically handles the setIsEditing(false) in `village-codes.tsx`
```
const closeForm = () => {
    setIsEditing(false);
    setFormData(DEFAULT_FORM);
  };
```

src/components/interactive/Modal.tsx
keiran0: Fixed

src/pages/settings/village-codes.tsx
keiran0: Fixed, good catch!

derrikzzz (approved):
This looks good to merge, thanks for working on this component haha will be helpful for reusability! 👍

PR #82, opened 2026-05-30
Toast dependency

keiran0:
This PR:
- Adds `react-hot-toast` as dependency
- Adds `<Toaster />` component to `_app.tsx`

### DIFF
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -5,6 +5,7 @@ import type { ReactElement, ReactNode } from "react";
 import { trpc } from "@/utils/trpc";
 import { useRouter } from "next/router";
 import withDefaultLayout from "@/components/layouts/withDefaultLayout";
+import { Toaster } from "react-hot-toast";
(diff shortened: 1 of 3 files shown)


src/pages/_app.tsx
```diff
@@ -36,10 +37,13 @@ function App({ Component, pageProps }: AppPropsWithLayout) {
...
-    <Component
-      {...pageProps}
-      key={isPatientRoute ? router.asPath : undefined}
-    />,
+    <>
+      <Component
+        {...pageProps}
+        key={isPatientRoute ? router.asPath : undefined}
+      />
+      <Toaster />
```
derrikzzz, line 45: Would it be better if we can add some configuration options here, such as:
```
<Toaster
  position="top-center"
  toastOptions={{ duration: 4000 }}
/>
```

derrikzzz (approved):
This looks good to merge, thanks for adding the toast notifications here!

PR #84, opened 2026-05-31
64 vitest unit testing

derrikzzz:
Closes #64 

## Description

- PR aims to introduce changes to set up Vitest as testing framework with React Testing Library. At the moment it only introduces unit testing on operations made in the village code page (page rendering, creation of village code form opening etc)

### Dependencies added
- Added testing-specific packages: vitest, @testing-library/react, @testing-library/jest-dom, @testing-library/user-event
- Added @vitejs/plugin-react and jsdom for React component testing support

### Documentation Updates
- Added testing section to README.md covering commands and writing patterns

### Test Coverage
- Only covers village codes page at the moment


package.json
```diff
@@ -11,7 +11,12 @@
     "prepare": "husky",
     "seed:db": "supabase db reset",
     "seed:users": "ts-node supabase/scripts/seed-users.ts",
-    "seed:all": "npm run seed:db && npm run seed:users"
+    "seed:all": "npm run seed:db && npm run seed:users",
+    "test": "vitest",
+    "test:ui": "vitest --ui",
```
keiran0, line 16: This line doesn't seem to work due to a missing dependency `@vitest/ui`.

package.json
```diff
@@ -11,7 +11,12 @@
...
+    "test:run": "vitest run",
+    "test:watch": "vitest --watch",
+    "test:coverage": "vitest run --coverage"
```
keiran0, line 19: This line doesn't seem to work due to a missing dependency `@vitest/coverage-v8`

package.json
```diff
@@ -38,21 +43,27 @@
...
     "eslint-config-next": "16.1.6",
     "husky": "^9.1.7",
+    "jsdom": "^29.1.1",
     "lint-staged": "^16.2.7",
     "prettier": "^3.8.0",
     "react-icons": "^5.5.0",
     "tailwindcss": "^4",
-    "typescript": "^5"
+    "typescript": "^5",
+    "vitest": "^1.6.1"
```
keiran0: The latest version of `vitest` is 4.1.8 (but just released yesterday, so perhaps we don't need to upgrade directly to that version specifically, due to risk of npm attacks)

We don't necessarily need to use the latest versions of dependencies, but perhaps it would be worth considering an upgrade to a more recent version, unless there's a reason why we should stick to 1.6.1

keiran0 (changes requested):

package.json
derrikzzz: Addressed this in commit [086d8d1](https://github.com/NUS-Project-SaBai/project-sabai/pull/84/commits/086d8d1183f688da1e89226f2ac0730397d8a968) of updating vitest version

keiran0 (changes requested):
Thanks for the dependency setup and for the village code tests! Just a few comments:

In the true implementation in `village_codes_router.ts`, whether hidden villages are included in the returned data is dependent on `includeHidden`, but the current `mockVillageCodes` used in all mock `list.useQuery` includes all village codes regardless of its `isVisible` attribute.

For the tests involving the hidden checkbox, it might be better to mock the same behaviour specified in the router (excluding CA when `includeHidden` is false, including CA otherwise). With the different behaviours in the mocked response, we can also test for correct village codes returned depending on checkbox state (displays hidden village codes only when hidden checkbox is checked for example) which might be a good testcase to have.

keiran0 (changes requested):
It might be better to make interactive elements like buttons `getByRole` instead, since `screen.getByText` or equivalent might have a 'false positive' causing the whole test to fail. For example in the test containing `user.click(screen.getByText("Cancel"))`, inserting a sentence with the word "Cancel" in it caused the test to break. Otherwise, LGTM (though might revisit when I've properly learnt vitest 😅)

keiran0 (approved):
LGTM!

PR #86, opened 2026-06-01
76 add use save on write hook

derrikzzz:
Closes #76 

## Description

- Adds a new custom hook useSaveOnWrite that automatically persists component state to localStorage with debounced writes.

## Changes made
- Added useSaveOnWrite hook
- Added helper function buildStorageKey (to generate unique storage keys)
- Updated new village code modal to use hook (persist across page refreshes)

### DIFF
--- a/src/pages/settings/village-codes.tsx
+++ b/src/pages/settings/village-codes.tsx
@@ -23,7 +23,6 @@ function VillageCodesPage() {
     "village-codes-form",
     DEFAULT_FORM,
     [], // No dependencies
-    500, // 500ms debounce
   );
 
   // 2. Data Fetching
--- a/src/pages/settings/village-codes.tsx
+++ b/src/pages/settings/village-codes.tsx
@@ -79,7 +79,6 @@ function VillageCodesPage() {
 
   const closeForm = () => {
     setIsEditing(false);
-    setFormData(DEFAULT_FORM);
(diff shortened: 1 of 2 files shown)


src/pages/settings/village-codes.tsx
```diff
@@ -72,7 +77,7 @@ function VillageCodesPage() {
 
   const closeForm = () => {
     setIsEditing(false);
-    setFormData(DEFAULT_FORM);
+    clearFormData();
```
keiran0: Would it be better to move this up to line 39? In that case, the values will still be retained if the user closes the modal without saving, but the values get saved and cleared if the user actually saves the data.

keiran0 (changes requested):
Just a small change required for the behaviour of the hook usage wrt to form submission vs modal closing, otherwise LGTM!

src/pages/settings/village-codes.tsx
derrikzzz: Have addressed in commit [dfc2672](https://github.com/NUS-Project-SaBai/project-sabai/pull/86/commits/dfc2672544b92a6d89bac18de6d533bc508b46d4)

keiran0 (approved):
LGTM, this hook will be very useful, thanks for the implementation!

PR #91, opened 2026-06-06
Facial recognition

### LINKED ISSUE #49: Add AWS Facial Recognition

keiran0:
This PR does 3 things: 
1) refactors the existing scanFace page in preparation for the actual facial recognition portion (closes #88)
- Adds a new component `RegistrationForm`
- Extracts some functions into external helper functions
- Adds the checkbox input type to `RHFInput`
- Refactors the `scanFace` page to use RHF, registrationForm and other reusable components
- Adds contactNo to the create procedure in patientsRouter
- Renames `scanFace` to `scan-face` as per path conventions

 2) the facial recognition portion itself (Closes #49)
- Modifies the patients schema in the database to accept a ~~`face_encoding`~~ `rekognition_face_id` column to store the encoding returned from AWS
- Creates some helper functions in `facialRecognition`
- Creates functions that makes use of AWS Rekognition client
- Implements the `findFaceMatches` and `listMatchingPatients` procedures in the `patientsRouter` that uses the above functions
- Modifies the `create` procedure in the router to write the return value from `generateFaceprint` to the database 

3) tests for the `scan-face` page 
- Adds the `name` attribute to RHFDropdown components for increased testability (unable to find by select/option types using `querySelector` unlike the other inputs)

This PR also changes some implementation in the patientsRouter in order to work with RHF and TRPC more effectively. The biggest change is `patientImage` (accepts a string instead of a file) and the conversion to file to upload to cloudinary is done on the server instead, with the base64 encoded string representation of the patient sent over the form. This might not be optimal especially if image sizes are large, but we can look into refactoring if this is an issue in the future. 

Note:
- The `actions` column in the found matches patients table is left purposely blank as it has to do with visits, which are out of scope from this PR. 
- `MatchingPatients` does not have success toasts because it's obvious to the user if things go right.
- `<Manual />` is a placeholder component meant to select a patient that have visited before, but the facial recognition was unable to pick them up. The intended implementation will be scanFace -> Manual mode -> user selects patient -> On selected patient, update faceprint and log a new visit. This will be implemented in a future PR, since this one is getting too large.
- Line 202 of `patients_router.ts` -> Left in a `console.log` here for easier testing. This line only targets the returned value from the API, but not the database - if you see something in the terminal here but not on the matched faces on the UI, it means that you need to register yourself as a patient first. This will be removed right before merging the PR.
- Dependencies were modified in this PR for AWS integration and to resolve other conflicts. The `pnpm-lock.yaml` file was erased and rebuilt from the `package.json` file.


src/components/scanFace/RegistrationPage.tsx
```diff
@@ -0,0 +1,159 @@
+import { useForm, FormProvider } from "react-hook-form";
+import { useState } from "react";
+import toast from "react-hot-toast";
+import { trpc } from "@/utils/trpc";
+import { RHFInput } from "@/components/interactive/RHF/RHFInput";
+import { RHFDropdown } from "@/components/interactive/RHF/RHFDropdown";
+import { Mode } from "@/pages/scanFace";
```
derrikzzz: There seems to be a circular page -> component dependency here, where `scanFace.tsx` imports both `MatchingPatients` and `RegistrationPage` components, and both imports `Mode` back from the `scanFace` page. Will it be better if we can avoid the circular dependency?

src/server/routers/patients_router.ts
```diff
@@ -24,6 +29,13 @@ const getPatientWithImageUrl = (patient: Patient) => ({
     : null,
 });
 
+function dataUrlToFile(dataUrl: string, filename: string): File {
+  const [header, base64] = dataUrl.split(",");
+  const mimeType = header.match(/:(.*?);/)?.[1] ?? "image/jpeg";
+  const bytes = toBytes(base64);
+  return new File([bytes], filename, { type: mimeType });
+}
+
```
derrikzzz: Would it be better if this helper sits inside `lib/utils/facialRecognition.ts` instead of being in this router?

src/db/schema.ts
```diff
@@ -77,6 +78,7 @@ export const patients = pgTable("patients", {
   hasBS2Card: boolean("has_bs2_card").notNull(),
   hasSabaiCard: boolean("has_sabai_card").notNull(),
   patientImagePublicId: text("patient_image_public_id").notNull(),
+  faceEncoding: text("face_encoding"),
```
derrikzzz: Minor nit (naming): should this be renamed to `rekognitionFaceId` for clarity?

src/lib/utils/facialRecognition.ts
```diff
@@ -0,0 +1,81 @@
...
+    CollectionId: env.COLLECTION_ID,
+    Image: image,
+  };
+
+  const client: RekognitionClient = new RekognitionClient();
+  const command = new IndexFacesCommand(input);
+
+  try {
+    const results: IndexFacesCommandOutput = await client.send(command);
+    return results.FaceRecords![0].Face?.FaceId;
```
derrikzzz: Will it be better to do 
```
// FaceRecords can be empty -> undefined
return results.FaceRecords?.[0]?.Face?.FaceId;
```
Rationale is, missing face becomes undefined instead of throwing

src/pages/scanFace.tsx
```diff
@@ -1,117 +1,25 @@
...
-    formData.append("contactNo", patientFormData.contactNo);
-    formData.append("gender", patientFormData.gender);
-    formData.append("drugAllergy", patientFormData.drugAllergy);
-    formData.append("dateOfBirth", patientFormData.dateOfBirth.toISOString());
-    formData.append("hasPoorCard", patientFormData.hasPoorCard.toString());
-    formData.append("hasBS2Card", patientFormData.hasBS2Card.toString());
-    formData.append("hasSabaiCard", patientFormData.hasSabaiCard.toString());
-    formData.append("villageCodeId", selectedVillageCodeId?.toString() ?? "");
-    formData.append("patientImage", patientImage); // ← File goes in last
+  const [mode, setMode] = useState<Mode>(Mode.REGISTERING); //should be mode.matching by default
```
derrikzzz: This may trip up developers who are reading the code and the comment, should this be changed to the code below to agree with the comment:
```
const [mode, setMode] = useState<Mode>(Mode.MATCHING);
```

src/server/routers/patients_router.ts
```diff
@@ -108,9 +117,19 @@ export const patientsRouter = router({
...
+      const patientImage: File = dataUrlToFile(
+        input.patientImage,
+        `${input.name}.jpg`,
+      );
+
       // Upload image to Cloudinary and get the public ID
-      const patientImagePublicId = await uploadToCloudinary(input.patientImage);
-      const newPatientInput = { ...patientData, patientImagePublicId };
+      const patientImagePublicId = await uploadToCloudinary(patientImage);
+      const faceEncoding = await generateFaceprint(input.patientImage);
```
derrikzzz: At the moment these are sequential awaits, but would it be better if we can wrap it with a `Promise.all`, something like
```
const [patientImagePublicId, faceEncoding] = await Promise.all([
  uploadToCloudinary(patientImage),
  generateFaceprint(input.patientImage),
]);
```
(not sure if this is considered premature optimization oops)

src/server/routers/patients_router.ts
```diff
@@ -85,16 +99,11 @@ export const patientsRouter = router({
...
-          z.enum(["true", "false"]).transform((val) => val === "true"),
-        ),
-        hasSabaiCard: zfd.text(
-          z.enum(["true", "false"]).transform((val) => val === "true"),
-        ),
-        patientImage: zfd.file(),
+        hasPoorCard: z.boolean(),
+        hasBS2Card: z.boolean(),
+        hasSabaiCard: z.boolean(),
+        patientImage: z.string(),
```
derrikzzz, line 98: Should there be validation over here for `patientImage`?

src/lib/utils/facialRecognition.ts
```diff
@@ -0,0 +1,81 @@
...
+ * @returns An array of FaceMatches.
+ */
+export async function searchFaceprint(str: string) {
+  const image: Image = toImage(str);
+  const input: SearchFacesByImageCommandInput = {
+    CollectionId: env.COLLECTION_ID,
+    Image: image,
+    // Keep the other attributes as default
+  };
+  const client: RekognitionClient = new RekognitionClient();
```
derrikzzz, line 48: Both `generateFaceprint` and `searchFaceprint` create a brand new
`RekognitionClient` every time they're called for each invocation from the looks of it here. Will it be better if we have something like a keep-alive connection to this client?

src/components/scanFace/MatchingPatients.tsx
```diff
@@ -0,0 +1,104 @@
...
+                <div className="text-sm font-medium text-slate-900">
+                  {patient.name}
+                </div>
+              </TableCell>
+              <TableCell>placeholder actions</TableCell>
+            </TableRow>
+          ))}
+        </tbody>
+      </table>
+      <button
```
derrikzzz: nit: The "Register new patient" button can be centered instead of left-aligned?

src/components/scanFace/RegistrationPage.tsx
```diff
@@ -0,0 +1,159 @@
...
+              type="submit"
+              disabled={isSubmitting}
+              className={`flex-1 px-4 py-2 rounded-lg font-medium ${isSubmitting ? "bg-gray-400 cursor-not-allowed" : "bg-indigo-600 text-white hover:bg-indigo-700 hover:cursor-pointer"}`}
+            >
+              Create New Patient
+            </button>
+          </div>
+        </form>
+      </FormProvider>
+      <button
```
derrikzzz: nit: would it be better if the "Find matches" button can be centered?

src/pages/scanFace.tsx
keiran0: I wrote that as a note to myself and resolved this but forgot to push the commit addressing my comment. Fixed!

src/components/scanFace/RegistrationPage.tsx
keiran0: Resolved

src/server/routers/patients_router.ts
keiran0: Fixed

src/lib/utils/facialRecognition.ts
keiran0: Fixed, good catch!

src/components/scanFace/RegistrationPage.tsx
keiran0: Oops, I missed that one. Fixed!

src/server/routers/patients_router.ts
keiran0: Fixed

src/server/routers/patients_router.ts
keiran0: Good catch, didn't register that both were sequential awaits. Not at all premature optimization since it's worth optimizing! Fixed

src/lib/utils/facialRecognition.ts
keiran0: Good news, it seems it's keep-alive by default in v3 of the sdk which we're currently on. https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/node-reusing-connections.html

However my current code explicitly starts up a new client for each function call so it doesn't make use of that. Good catch on the optimization aspect! I have created a new issue in #109 to prevent this PR from getting more bloated.

src/db/schema.ts
keiran0: Fixed

src/server/routers/patients_router.ts
```diff
@@ -184,4 +194,77 @@ export const patientsRouter = router({
...
       return { success: !!result };
     }),
+
+  // find all matches
+  findFaceMatches: protectedProcedure
+    .input(z.object({ picture: z.string() }))
+    .mutation(async ({ input }) => {
+      try {
+        const res = await searchFaceprint(input.picture);
+        console.log("result", res);
```
derrikzzz: Should we remove this console.log statement!

src/db/schema.ts
```diff
@@ -61,6 +61,7 @@ Patients Table:
 - has_bs2_card: Indicates if the patient has a BS2 card or not.
 - has_sabai_card: Indicates if the patient has a Sabai card or not.
 - patientImagePublicId: Cloudinary public ID of the patient's image.
+- face_encoding: Face encoding for facial recognition.
```
derrikzzz: Should the comment be updated here to reflect the new naming from `face_encoding` to `rekognitionFaceId`?

keiran0: Fixed, as well as the others who use snake case instead of camel case.

src/server/routers/patients_router.ts
keiran0: Completed the changes, removed!

src/components/scanFace/MatchingPatients.tsx
```diff
@@ -0,0 +1,121 @@
...
+export default function MatchingPatients({
+  imgDetails,
+  setMode,
+}: {
+  imgDetails: string | null;
+  setMode: React.Dispatch<React.SetStateAction<Mode>>;
+}) {
+  const findFaceMatchMutation = trpc.patientsRouter.findFaceMatches.useMutation(
+    {
+      onSuccess(test) {
```
derrikzzz: nit: will it be better if `test` is renamed to `response`?

keiran0: It definitely would, a placeholder name made it into my PR. Fixed

derrikzzz (approved):
Looks good to merge, thanks for working on this facial recognition feature!

Let's try to make sure that the PRs in the future are not too long too haha! 😆

PR #108, opened 2026-06-14
107 fix image capture resize

derrikzzz:
Closes #107 

## Description 
1. WebcamInput responsive fix — replaced fixed width/height pixel props with a size prop that acts as a
maxWidth. The webcam feed now uses `width="100%"` and the preview container uses `aspect-square`, so both
shrink to fit narrow mobile viewports instead of overflowing and causing a layout shift when the
camera opens.

2. VillageSelector centering — restructured the mobile navbar to `grid-cols-3`, so the selector sits in the true center column between the logo and menu button.

## Screenshot / Recording to reflect changes

https://github.com/user-attachments/assets/e6095339-7e58-4293-bd4d-6cbe36927352

### DIFF
--- a/src/components/layouts/SidebarLayout.tsx
+++ b/src/components/layouts/SidebarLayout.tsx
@@ -23,24 +23,26 @@ export default function SidebarLayout({ children }: SidebarLayoutProps) {
   return (
     <div className="flex h-screen flex-col sm:flex-row">
       {/* mobile sidebar */}
-      <div className="grid sm:hidden grid-cols-3 w-full p-2 items-center bg-[var(--color-navbar)]">
-        <SabaiLogo />
-        <div className="flex justify-center">
-          <VillageSelector />
+      <div className="sm:hidden relative">
+        <div className="grid grid-cols-3 w-full p-2 items-center bg-[var(--color-navbar)]">
+          <SabaiLogo />
+          <div className="flex justify-center">
+            <VillageSelector />
+          </div>
+          <div className="flex justify-end">
+            <button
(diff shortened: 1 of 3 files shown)


src/components/layouts/SidebarLayout.tsx
```diff
@@ -23,25 +23,27 @@ export default function SidebarLayout({ children }: SidebarLayoutProps) {
   return (
     <div className="flex h-screen flex-col sm:flex-row">
       {/* mobile sidebar */}
-      <div className="grid sm:hidden grid-cols-3 w-full p-2 items-center bg-[var(--color-navbar)]">
-        <SabaiLogo />
-        <div className="flex justify-center">
-          <VillageSelector />
+      <div className="sm:hidden relative">
```
keiran0: Would it be better to include `useClickOutside` for the mobile sidebar?

````
const dropdownRef = useClickOutside<HTMLDivElement>(() => setMobileSidebarOpen(false));
...
<div className="sm:hidden relative" ref={dropdownRef}>
....
```

keiran0 (changes requested):
I think it might be good to use the custom hook so that the sidebar is collapsed when the user wants to fill out the form or interact with the `scanFace` page, otherwise LGTM and thanks for catching this bug!

src/components/layouts/SidebarLayout.tsx
derrikzzz: Have added hook in [2c13ac6](https://github.com/NUS-Project-SaBai/project-sabai/pull/108/commits/2c13ac6af8338f910a4197ef9f216ec446e13c70)

keiran0 (approved):
LGTM!

PR #112, opened 2026-06-19
110 improve vision page layout

derrikzzz:
Closes #110 

## Description

- Improves the Vision page layout and makes the patient list responsive across mobile, tablet, and desktop. 
- Also adds a shared `formatPatientId` / `formatPatientCode` helper 
- Made change to include latest-visit village in the patient list query

## Testing
- Manually verify responsive layouts at mobile / tablet / desktop breakpoints
---

**Mobile View:**

Screenshot 2026-06-19 at 8 10 59 PM

---

**Tablet View:**
 
Screenshot 2026-06-19 at 8 47 43 PM

---

**Desktop View:**
 
Screenshot 2026-06-19 at 8 47 53 PM

---

### DIFF
--- a/src/server/routers/patients_router.ts
+++ b/src/server/routers/patients_router.ts
@@ -3,10 +3,10 @@ import { zfd } from "zod-form-data";
 import { uploadToCloudinary } from "@/server/utils/cloudinary";
 import { router, protectedProcedure } from "@/server/trpc";
 import { db } from "@/db/drizzle";
-import { patients, genderEnum, Patient, visits } from "@/db/schema";
+import { patients, genderEnum, visits, villageCodes } from "@/db/schema";
 import serverEnv from "@/lib/envVariables";
 import { TRPCError } from "@trpc/server";
-import { eq, inArray } from "drizzle-orm";
+import { eq, desc, inArray } from "drizzle-orm";
 import {
   generateFaceprint,
   searchFaceprint,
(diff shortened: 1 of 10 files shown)


src/server/routers/patients_router.ts
```diff
@@ -32,6 +36,16 @@ const getPatientWithImageUrl = (patient: Patient) => ({
 export const patientsRouter = router({
   // List all patients with village details
   list: protectedProcedure.query(async () => {
```
keiran0, line 38: Would it be better to create a new procedure called `listWithLatestVisit` or something along those lines? This procedure name and the return value containing the village code seems to suggest to imply that patients are intrinsically tied to a village. The `list` procedure can be left alone for the patients page (view all patients independent of their visits)

keiran0 (changes requested):
Just a small review for now, feel free to disregard and batch address it together with my later reviews instead!

src/server/routers/patients_router.ts
derrikzzz: I think that this cleanup should be made in a separate PR with a description of why the split matters, but this is a good catch, do agree that it should be captured separately!

keiran0: Okay, will defer until the next PR!

keiran0 (approved):
LGTM for this PR, will await your next PR addressing the current review!

PR #116, opened 2026-06-26
Vision form

derrikzzz:
## Description

- Implements the eyesight prescription form on the update glasses page.
- Tests to be made in separate PR, LOC reaching almost 1k

## Changes made
- `formatVisitDate` util: Extracted from vitals page into `src/lib/utils/visit.ts` and reused across both pages.
- new `EyesightForm` component added.
- Update glasses page: Replaced the placeholder with the live EyesightForm and typed the shared form.
- DB: Added `left_prescribed_glasses_degree` and `right_prescribed_glasses_degree` columns to the eyesight table with a migration
- Router: Wired up the two new fields into selectEyesightFields and createEyesightInput in `eyesight_router.ts`

## Screenshot of UI

Screenshot 2026-06-26 at 8 44 38 AM

### DIFF
--- a/src/components/vision/EyesightForm.tsx
+++ b/src/components/vision/EyesightForm.tsx
@@ -17,10 +17,8 @@ export type EyesightFormValues = {
   leftPrescribedGlassesDegree?: string | null;
   rightPrescribedGlassesDegree?: string | null;
   comments?: string | null;
-  visitSelect: string;
 };
 
-/** Empty form state used as the baseline for a visit with no record yet. */
 const BLANK_EYESIGHT: EyesightFormValues = {
   leftEyeDegree: "",
   rightEyeDegree: "",
@@ -31,12 +29,8 @@ const BLANK_EYESIGHT: EyesightFormValues = {
   leftPrescribedGlassesDegree: "",
   rightPrescribedGlassesDegree: "",
   comments: "",
-  visitSelect: "",
 };
 
-/**
- * Left/right input groups rendered in the prescription form, declared once here.
- */
 const EYE_SECTIONS: {
   title: string;
   fields: { name: keyof EyesightFormValues; label: string }[];
@@ -77,20 +71,17 @@ const EYE_SECTIONS: {
   },
 ];
 
-/**
- * Prescription form for a single visit. Loads any existing eyesight record and
- * upserts (create or insert) on save (create when none exists, otherwise will update by visit).
- */
 export function EyesightForm({ visitId }: { visitId: number }) {
   const {
     reset,
+    getValues,
     handleSubmit,
     formState: { isDirty },
-  } = useFormContext<EyesightFormValues>();
+  } = useFormContext<EyesightFormValues & { visitSelect: string }>();
 
   const { data: eyesightData, isLoading: eyesightLoading } =
     trpc.eyesightRouter.getByVisitId.useQuery(
-      { visitId: visitId },
+      { visitId },
       { enabled: !!visitId },
     );
 
@@ -113,10 +104,11 @@ export function EyesightForm({ visitId }: { visitId: number }) {
 
   useEffect(() => {
     if (eyesightLoading) return;
+    // Preserve visitSelect — it's owned by the parent page, not this form
     reset({
       ...BLANK_EYESIGHT,
       ...eyesightData,
-      visitSelect: visitId.toString(),
+      visitSelect: getValues("visitSelect"),
     });
   }, [eyesightData, eyesightLoading, reset, visitId]);
 
@@ -129,7 +121,7 @@ export function EyesightForm({ visitId }: { visitId: number }) {
     }
 
     const payload = {
-      visitId: visitId,
+      visitId,
       leftEyeDegree: data.leftEyeDegree || undefined,
       rightEyeDegree: data.rightEyeDegree || undefined,
       leftEyePinhole: data.leftEyePinhole || undefined,
--- a/src/components/vision/EyesightForm.tsx
+++ b/src/components/vision/EyesightForm.tsx
@@ -83,7 +83,7 @@ export function EyesightForm({
     formState: { isDirty },
     reset: resetCtx,
   } = useFormContext<EyesightFormValues>();
-  
+
   const reset = resetCtx as (
     values: EyesightFormValues & { visitSelect: string },
   ) => void;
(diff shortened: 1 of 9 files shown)


src/pages/vision/update-glasses/[patientId].tsx
```diff
@@ -71,57 +72,63 @@ function UpdateGlassesPage() {
...
+                      name="visitSelect"
+                      label="Choose visit"
+                      dropdownOptions={visits.map((visit) => ({
+                        label: formatVisitDate(visit.date),
+                        value: visit.id.toString(),
+                      }))}
+                    />
+                  </div>
+                  <div className="text-center text-sm text-slate-600">
+                    Filling up vision form for visit on{" "}
```
Comment (keiran0) [inline src/pages/vision/update-glasses/[patientId].tsx:104]: There is some inconsistency with the styling here, compared to the vitals form page which are both very similar. Would it be better to standardize both pages, particularly this line so that users can expect to see the same format for both stations?

src/pages/vision/update-glasses/[patientId].tsx
Comment (keiran0) [inline src/pages/vision/update-glasses/[patientId].tsx]: In following the change from `-` to `—`, should the breadcrumb also be updated for consistency?

src/components/vision/EyesightForm.tsx
```diff
@@ -0,0 +1,186 @@
...
+  const createEyesightMutation = trpc.eyesightRouter.create.useMutation();
+  const updateEyesightMutation =
+    trpc.eyesightRouter.updateByVisitId.useMutation();
+
+  useEffect(() => {
+    if (eyesightLoading) return;
+    reset({ ...BLANK_EYESIGHT, ...eyesightData, visitSelect });
+  }, [eyesightData, eyesightLoading, reset, visitId, visitSelect]);
+
+  if (eyesightLoading) return <LoadingSpinner message="Loading vision data" />;
```
keiran0, line 108: Should the return statement be formatted with curly braces on the next line for readability?

src/components/vision/EyesightForm.tsx
naa-siuuuu-ff, line 1: Think in the future, might refactor the vitals form into components also like this

naa-siuuuu-ff (approved):
LGTM! Just some merge conflict with import statements

src/pages/vision/update-glasses/[patientId].tsx
```diff
@@ -29,37 +34,33 @@ function UpdateGlassesPage() {
     name: "visitSelect",
   });
 
-  if (patientLoading || visitsLoading) {
+  // Automatically select visit if only got one
+  useEffect(() => {
+    if (visits && visits.length === 1) {
+      methods.setValue("visitSelect", visits[0].id.toString());
+    }
+  }, [visits, methods]);
```
Comment (keiran0) [inline src/pages/vision/update-glasses/[patientId].tsx]: Adding 'methods' here was probably to silence the linter, but there's another workaround. Would it be better to destructure methods and use `setValue` in the dependency array so that it's stable?

As a nice side effect (no pun intended), it also is compatible with a planned future release of RHF.

See documentation: https://react-hook-form.com/docs/useform (search 'useForm return and useEffect dependencies')
> In a future major release, useForm return will be memoized to optimize performance and reflect changes in formState. As a result, adding the entire return value of useForm to a useEffect dependency list may lead to infinite loops.
> The recommended way is to pass destructured methods to the dependencies of an useEffect

keiran0 (changes requested):
Hold off on merging, some small changes requested from me. Mostly nitpicks with consistency, but looks good otherwise! Also seconding the point about resolving the merge conflict.

src/pages/vision/update-glasses/[patientId].tsx
Comment (derrikzzz) [reply src/pages/vision/update-glasses/[patientId].tsx:104]: Oh actually do you mean the visit dropdown here (Screenshot attached):

Screenshot 2026-06-29 at 8 31 54 PM

If yes, then am proposing a different style for it, rationale is that user can see what visit they selected next to dropdown without having to scroll to the vitals form area component.

Another rationale can be that this can arguably be a cleaner separation? The header can stay focused on which visit they are working on while the form body is more focused on filling up of entries

src/components/vision/EyesightForm.tsx
derrikzzz: For this, should we defer to another issue? Because there will be other files to modify such as in `src/pages/ivtals/[id]` with:
```
if (vitalsLoading) return <LoadingSpinner message="Loading Vitals" />;
```

src/pages/vision/update-glasses/[patientId].tsx
Comment (derrikzzz) [reply src/pages/vision/update-glasses/[patientId].tsx]: Changed to using emdash in [0f31d2e](https://github.com/NUS-Project-SaBai/project-sabai/pull/116/commits/0f31d2e2195b8d3e3a894df82c3b6144d2b01b52)

src/pages/vision/update-glasses/[patientId].tsx
Comment (derrikzzz) [reply src/pages/vision/update-glasses/[patientId].tsx]: Destructured `setValue` from methods in [58b548b](https://github.com/NUS-Project-SaBai/project-sabai/pull/116/commits/58b548bbeab39fd4b853319930d453ed84f2b078)

src/pages/vision/update-glasses/[patientId].tsx
Comment (keiran0) [reply src/pages/vision/update-glasses/[patientId].tsx:104]: Oh yes, visit dropdown and the position of the "filling up..." indicator on the right hand side. It actually kind of seems a bit redundant since the dropdown already implies that the vision form belongs to 21 May 2026 17:30's visit.

On the vitals form, that indicator is at the bottom of the dropdown instead of the side.

src/components/vision/EyesightForm.tsx
```diff
@@ -0,0 +1,186 @@
...
+  const utils = trpc.useUtils();
+
+  const createEyesightMutation = trpc.eyesightRouter.create.useMutation();
+  const updateEyesightMutation =
+    trpc.eyesightRouter.updateByVisitId.useMutation();
+
+  useEffect(() => {
+    if (eyesightLoading) return;
+    reset({ ...BLANK_EYESIGHT, ...eyesightData, visitSelect });
+  }, [eyesightData, eyesightLoading, reset, visitId, visitSelect]);
```
keiran0: It appears as if some variables here are redundant (though i see the intent in putting them here to silence the linter). Would a comment help with justifying the variables here?

keiran0 (commented):
Just a small add-on where I think a comment might be useful

src/pages/vision/update-glasses/[patientId].tsx
Comment (derrikzzz) [reply src/pages/vision/update-glasses/[patientId].tsx:104]: Oh I get what you mean, maybe we can defer it to another PR, since this is more of a UX issue?

keiran0 (approved):
LGTM, reminder to create the issues addressing the reviews later on!

PR #118, opened 2026-06-26
Refactor village codes

### LINKED ISSUE #31: Village Code Page

### LINKED ISSUE #79: Refactor and add error handling to village-codes page

keiran0:
Closes #95, closes #31, closes #79

## Description

This PR refactors the village code to use the same practices as the rest of the pages in the app.

Changes made (directly related to feature)
- Addedtoasts when a village code is modified, created, or deleted 
- Used RHF components and their associated hooks instead of plain HTML components with `setState` 
- Added error handling (toast + console.error, placeholder until we actually implement logging)
- Used`useSaveOnWrite` hook on village code creation to save the entire form to local storage
- Changed deletion modal from `confirm` to an actual component and form.
- Added testcases

Changes made (pre-requisite for feature or other trivial misc fixes):
- Addedcolor and checkbox types to `RHFInput` components
- Fixed typo in `medication-active-ingredients` testcase
- Added missing 'code' attribute in village code router update procedure
- Extracted settings into a placeholder `index.tsx`, village codes page now lives in its own folder with all its associated components and types. The modals were moved out of the main file because they're big components that would be re-defined on every state change

Things to note:
- If the routing change resulted in any issues, clear `.next/` then relaunch the app, potential issues can be caused by the nextjs cache


noahseethorcodes:
Quick smoke test:
1. Go to page (page loads, working fine)
2. Click "New Village Code" (modal pops up, input works)
3. Click save -> code doesn't get saved
4. Reload -> page breaks

Can help me sanity check if the behaviour is the same on your machine?
UPDATE: runtime error caused by next caching issue. fixed by clearing .next/ folder locally. save mechanic still broken

https://github.com/user-attachments/assets/a2288d23-3f11-40ea-b4d7-201b4acd5ca1

keiran0:
Reverting commit because somehow the button refactoring broke stuff

src/pages/settings/village-codes/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+              Show Hidden
+            </label>
+            <Button
+              onClick={() => setMode("editing")}
+              colour="emerald"
+              title="New Village Code"
+            />
+          </div>
+        </div>
+        <div className="min-h-screen bg-slate-50 p-8">
```
noahseethorcodes: is this `div` and its attributes intentionally repeated from [L103](https://github.com/NUS-Project-SaBai/project-sabai/pull/118/changes#diff-ed81e8dfc1c18660c1bd4e56c7dc6270407086e5c26fd9e3898376eb6a210ab2R103)?

src/components/village-codes/ChangeModal.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+  const updateMutation = trpc.villageCodesRouter.update.useMutation({
+    onSuccess: () => {
+      utils.villageCodesRouter.list.invalidate();
+      form.reset();
+      onClose();
+      toast.success("Village code updated!");
+    },
+    onError: (err) => {
+      console.log(err);
```
noahseethorcodes: should be `console.error` instead of `console.log`?

src/components/village-codes/DeleteModal.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+      <div className="flex gap-2 justify-center">
+        <Button
+          onClick={onClose}
+          colour="red"
+          title="Cancel"
+          className="w-full"
+        />
+        <Button
+          onClick={() => deleteMutation.mutate({ id: activeForm!.id! })}
```
noahseethorcodes, line 45: `activeForm` prop typed `FormFields | null`. No guard before `activeForm!.id!`. If `DeleteModal` renders with `activeForm=null` this crashes.

src/components/village-codes/ChangeModal.tsx
```diff
@@ -0,0 +1,187 @@
...
+    if (activeForm && activeForm.id) {
+      handleEdit(data);
+    } else {
+      createMutation.mutate(data);
+    }
+  };
+
+  const handleEdit = async (data: FormFields) => {
+    const numDirtyFields = Object.keys(form.formState.dirtyFields).length;
+    if (form.formState.isDirty || numDirtyFields > 0) {
```
noahseethorcodes: `isDirty` already reflects dirty fields count. Second condition is dead code.

```suggestion
    if (form.formState.isDirty) {
```

src/lib/utils/villageCodeTypes.ts
```diff
@@ -0,0 +1,16 @@
...
+  colorHex: "#3b82f6", // Default blue
+  isVisible: true,
+};
+
+export type FormFields = {
+  id?: number;
+  name: string;
+  code: string;
+  colorHex: string;
+  isVisible: boolean | undefined;
```
noahseethorcodes, line 15: can consider providing a default value (e.g isVisible defaults to `true`) instead of allowing `undefined`

src/__tests__/pages/settings/village-codes/index.test.tsx
```diff
@@ -180,6 +185,51 @@ describe("VillageCodesPage", () => {
...
+    });
+
+    const codeInput = document.querySelector('input[name="code"]');
+    expect(codeInput).toBeInTheDocument();
+    await user.type(codeInput!, "1111");
+
+    expect(screen.queryByRole("status")).not.toBeInTheDocument();
+    await user.click(screen.getByRole("button", { name: "Save" }));
+
+    const toast = screen.getByRole("status");
```
noahseethorcodes: `import { toast, Toaster }` at top, then `const toast = screen.getByRole("status")` shadows imported toast. (see [variable shadowing](https://www.geeksforgeeks.org/javascript/variable-shadowing-in-javascript/))

```suggestion
    const toastEl = screen.getByRole("status");
```

src/components/village-codes/ChangeModal.tsx
```diff
@@ -0,0 +1,187 @@
...
+  });
+
+  const watchedColor = useWatch({
+    name: "colorHex",
+    compute: (data: string) => {
+      return data;
+    },
+    defaultValue: DEFAULT_FORM.colorHex,
+    control: form.control,
+  });
```
noahseethorcodes, line 52: `compute: (data) => data` — identity no-op. Was compute meant to transform the watched value somehow? 
If not, just 
```suggestion
  const watchedColor = useWatch({
    name: "colorHex",
    defaultValue: DEFAULT_FORM.colorHex,
    control: form.control,
  });
```

src/components/interactive/RHF/RHFInput.tsx
```diff
@@ -74,7 +81,7 @@ export function RHFInput({
         {...registerProps}
         {...props}
         className={clsx([
-          "border border-gray-300 rounded bg-white px-3 py-2 transition focus:outline-none focus:ring-2 focus:ring-blue-400 focus:border-blue-400",
+          "border border-gray-300 rounded bg-white px-3 py-2 transition focus:outline-none focus:ring-2 focus:ring-blue-400 focus:border-blue-400 h-10",
```
noahseethorcodes: `h-10` applied globally. Checkbox type gets forced height, which may break its layout (checkboxes are typically inline and don't need/want a fixed block height). Consider making it conditional on `type !== "checkbox"`.

noahseethorcodes (changes requested):
first pass of review comments. apologies if this is just bad code I wrote in previously and now you have to fix 😅

src/components/village-codes/ChangeModal.tsx
keiran0: Fixed 30b7320

src/components/village-codes/ChangeModal.tsx
keiran0: Compute was previously logged to console to make sure what I was doing was correct, but I forgot to remove it, fixed in 3cdacb6

src/__tests__/pages/settings/village-codes/index.test.tsx
keiran0: Good catch, didn't realise until you mentioned it, fixed in 437617c

src/components/interactive/RHF/RHFInput.tsx
keiran0: I've made it conditional on the colour rhfinput type instead (matches more with my original intentions for adding the h-10, in 6724c53

src/pages/settings/village-codes/index.tsx
keiran0: Fixed in 33c5e6b, good catch, I didn't even notice the strangely spaced UI until you pointed it out hahaha

src/components/village-codes/DeleteModal.tsx
keiran0: Fixed in e02edd7, thanks!

src/components/village-codes/ChangeModal.tsx
keiran0: Fixed in b8ace64

src/lib/utils/villageCodeTypes.ts
keiran0: `isVisible`'s value is restricted by the database, and the provided default value. I tried changing this line in `villageCodeTypes.ts`, but then it causes more type errors (tracing it to `NewVillageCode`). Turns out, `notNull` with a default value provided will cause the field to accept undefined values.

The type is constrained by both the schema and the default form, so perhaps it's acceptable to leave this line as is. Alternatively, perhaps we can change the schema to not have a default value and solely rely on the form's default value.

noahseethorcodes: understood. thanks for clarifying!

noahseethorcodes (approved):
LGTM! Thanks for making the changes and looking into the query for the default value on `isVisible`. Can consider re-working the changes for that in a future PR. Approving this first.

PR #119, opened 2026-06-28
117 refactor patient list procedure

derrikzzz:
Closes #117

## Description:

- `patientsRouter.list` now returns plain patient rows with no JOINS — temporarily used by patients page 
- New `patientsRouter.listWithLatestVisit` carries the existing visits/villageCodes JOIN — used by the vision page which for now needs villageCode and villageColorHex, whether this will be used in the other pages will be in future works
- Updated `trpc-types` to export types for both procedures
- Vision page updated to use `listWithLatestVisit` and its corresponding type
- Delete `src/__tests__/pages/vitals/[id]/index.test.tsx` file, tests for vitals will be covered in separate PR as this PR is focused on refactoring

### DIFF
--- a/src/pages/vision/index.tsx
+++ b/src/pages/vision/index.tsx
@@ -10,7 +10,7 @@ import TableRow from "@/components/TableRow";
 import TableCell from "@/components/TableCell";
 import { ReactNode } from "react";
 import { formatPatientCode } from "@/lib/utils/patient";
-import { PatientsRouterListItem } from "@/utils/trpc-types";
+import { PatientsRouterListWithLatestVisitItem } from "@/utils/trpc-types";
 
 /**
  * Renders a patient's code with a small dot in the village colour beside it.
@@ -62,7 +62,7 @@ function PatientCard({
   patient,
   onUpdateGlasses,
 }: {
-  patient: PatientsRouterListItem;
+  patient: PatientsRouterListWithLatestVisitItem;
   onUpdateGlasses: (patientId: number) => void;
 }) {
   return (
@@ -94,7 +94,7 @@ function PatientTable({
   patients,
   onUpdateGlasses,
 }: {
-  patients: PatientsRouterListItem[];
+  patients: PatientsRouterListWithLatestVisitItem[];
   onUpdateGlasses: (patientId: number) => void;
 }) {
   return (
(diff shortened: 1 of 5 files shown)


src/pages/vision/index.tsx
```diff
@@ -161,7 +161,7 @@ function VisionPage() {
     data: patients,
     isLoading,
     isError,
-  } = trpc.patientsRouter.list.useQuery();
```
keiran0, line 164: The vitals page currently doesn't support showing the village code of the latest visit, but should the same modification be also made with the vitals page so that the visit (latest on top) ordering is also reflected?

keiran0 (changes requested):
Just a minor change requested for the vitals page that is relevant to this PR, otherwise LGTM!

src/pages/vision/index.tsx
derrikzzz: Whoops, addressed this in commit [0e9bbb4](https://github.com/NUS-Project-SaBai/project-sabai/pull/119/commits/0e9bbb4c18f099b1d1c209084ff2ae979bbd79ac) on ordering latest visit at top by dates

src/server/routers/patients_router.ts
```diff
@@ -87,7 +88,9 @@ export const patientsRouter = router({
       })
       .from(patients)
       .leftJoin(latestVisit, eq(latestVisit.patientId, patients.id))
-      .leftJoin(villageCodes, eq(villageCodes.id, latestVisit.villageCodeId));
+      .leftJoin(villageCodes, eq(villageCodes.id, latestVisit.villageCodeId))
```
keiran0, line 91: I'm not sure if this is intended behaviour or if this is something that won't happen in the actual usecase so there's no need to account for it, (considering if a patient is created during registration, a visit will automatically be created for them, but I'm also hesitant to enforce this in case there's any case where a patient might register but not actually visit tho that seems unlikely), but in the current local dev database (seeded by scripts), there are patients who are registered who do not have visits.

For those patients, they are displayed right at the very top of the list because of the left join. Should we consider changing implementation so that patients without any visits are displayed at the bottom so that this procedure doesn't hide the patients who are actually going to be screened soon? Alternatively, we can consider omitting the patients without visits by joining based on the visits table instead of the patients table, then using an inner join for the patients.

keiran0 (changes requested):
Just some changes to the router procedure, otherwise LGTM

keiran0 (approved):

PR #123, opened 2026-06-28
122 replace ts-node with tsx and fix seed:all to use pnpm

derrikzzz:
Closes #122 

## Description 

- Replaced with tsx to handle ESM out of the box
- Fixed `seed:all` script to use pnpm instead of npm run for consistency with package manager used for this project


noahseethorcodes (approved):
lgtm. thanks for cleaning up my mess HAHA

PR #125, opened 2026-06-30
Refactor button

keiran0:
closes #124 

## Description 

Previously, some buttons didn't use the custom Button component despite potentially benefiting from it, due to some colours being missing. 

This PR:

- Adds `red` and `indigo` styles to `Button` component
- Adds `className` prop to `Button` component
- Refactors eligible buttons to use the component
- Replaces `OldButton` usage with `Button`
- Removes `OldButton` 
- Changes the default variant prop of the component to filled (all usages are filled anyway) and deduplicated usages of the component


src/components/interactive/Button/buttonStyles.ts
```diff
@@ -42,6 +42,18 @@ const ButtonTWClassName: Record<
     text: "text-emerald-600 hover:text-emerald-700",
     icon: "text-emerald-600 hover:text-emerald-700",
   },
+  red: {
+    filled: "bg-red-700 text-white hover:bg-red-800",
+    outline: "border border-red-700 text-red-700 hover:bg-red-800",
+    text: "text-red-700 hover:text-red-700",
+    icon: "text-red-700 hover:text-red-700",
```
derrikzzz: Since `text-red-700` is on both sides, will this be a no-op here?

src/components/scanFace/RegistrationPage.tsx
```diff
@@ -131,30 +131,27 @@ export default function RegistrationPage({
...
+        <Button
           onClick={() => setMode(Mode.MATCHING)}
-          className="flex-1 px-4 py-2 rounded-lg font-medium bg-indigo-700 text-white"
-        >
-          Match Instead
-        </button>
+          colour="indigo"
+          title="Match Instead"
+          className="w-full"
+        />
```
derrikzzz, line 152: Instead of these buttons taking the full width, would it be better if we remove the `w-full` from both buttons here and add `justify-center` to the div wrapper of the buttons to center them?

src/components/interactive/Button/buttonStyles.ts
keiran0: Whoops typo, fixed

src/components/scanFace/RegistrationPage.tsx
keiran0: That does look nicer, fixed

derrikzzz (approved):
Looks good to merge! Thanks for working on this Button component to be used in other places!

PR #101, opened 2026-06-09
85 village code local storage visits

derrikzzz:
Closes #85 

## Description
- Implements localStorage persistence for village code selection across browser sessions
- Adds village code integration to patient creation workflow
- Provides persistent village selector UI in navigation

## User Flow for this feature
  1. Medical student / Doctor selects village code from navigation dropdown
  2. Selection persists in localStorage across browser sessions
  3. When creating new patients, selected village code automatically applies
  4. Visit records created with current village code + timestamp (currently shown in patient detail page, but future PR for vitals should take note that there should be a dropdown for visits in the vitals page for each patient @naa-siuuuu-ff)
  5. Medical student / Doctor can change village selection mid-day for different locations

 ##  Test Plan
- Village code selection saves to localStorage
- Patient creation flow should now include village code
- Visit auto-creation works with selected village

PR #126, opened 2026-06-30
Refactor loading spinner

keiran0:
Closes #106 

## Description

Previously, the `LoadingSpinner` component came with a `p-12` baked into the div. This PR:
- Refactors the component to use no padding by default
- Adds a `className` prop that is used by the spinner 
- Refactors usages of the component to add on the padding where necessary

This addresses the comment in PR #101 as well.


derrikzzz (approved):
Looks good to merge in, thanks for refactor to include optional className prop to the `LoadingSpinner` component!

PR #127, opened 2026-06-30
Add tanstack table and other chores

keiran0:
## Details

It would be useful to have a filterable/sortable/groupable table, especially for the pharmacy stock (grouping by active ingredients, sorting by expiry date, filtering by stock status etc). Doing this from scratch might require a lot of work and there's no need to re-invent the wheel. Tanstack table can help with this by having these features out of the box. In addition, we already have our current compatible table helper components (other libraries with similar feature are shadcn's table component which also uses tanstack table anyway, but we have our own reusable components already), ag-grid (overkill and grouping feature not in their community edition) and others with similar features.

This PR also removes the pnpm overrides from package.json as it no longer reads the overrides from there. It has been shifted to pnpm-workspace.yaml (See: https://pnpm.io/migration for more detail about pnpm v11).

Finally, it also adds 3 packages to allow builds since these are all expected packages.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -1,5 +1,9 @@
 packages:
   - .
-ignoredBuiltDependencies:
-  - sharp
-  - unrs-resolver
+allowBuilds:
+  esbuild: true
+  sharp: true
+  unrs-resolver: true
+overrides:
+  "@types/react": "19.2.11"
+  "@types/react-dom": "19.2.3"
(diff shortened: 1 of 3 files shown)


pnpm-workspace.yaml
```diff
@@ -1,5 +1,9 @@
 packages:
   - .
-ignoredBuiltDependencies:
-  - sharp
-  - unrs-resolver
+allowBuilds:
+  esbuild: true
```
derrikzzz, line 4: Is the inclusion of `esbuild` here intentional?

keiran0: Oh yes, forgot to mention in my PR description, edited!

derrikzzz (approved):
Looks good to merge in, will be curious on what are common operations by them for the pharmacy stock!

PR #130, opened 2026-07-01
Medication stock

keiran0:
## Description

This PR:
- Adds some methods to the medication stock and medication brands router to display names of associated active ingredients
- Remove the placeholder ids in the medication stock page table and replaces them with the actual names of the brands and active ingredient
- Adds a modal and a form for users to add new stock. 
- In the modal: The dropdown options in the stock status field corresponds to the stockStatus enum as defined by the db schema (though more states can and should be added in the future)
- In the modal: The dropdown options in the brand + active ingredient field corresponds to the brands currently in the database. However, more work needs to be done to make it searchable, though that's a future consideration.
- Adds a partial implementation of the split stock functionality, to be fleshed out and properly written in another PR. The only thing the confirmation button does is print the splits to console.
- Adds a placeholder implementation of the edit action. All it does is just display 'edit' on the screen, to be fleshed out in another PR.

Future considerations:
- Disposed/dispensed/donated etc should not be included in the table by default. 
- Medication stock qty with exactly the same (other) fields should be grouped together (sum+groupby), but this will also be implemented after the `remarks` column has been implemented.


src/lib/utils/medication-stock.ts
```diff
@@ -0,0 +1,26 @@
+import { MedicationStock } from "@/db/schema";
+
+export enum StockStatus {
+  ACTIVE = "active",
+  DISPOSED = "disposed",
+  DONATED = "donated",
+  EXPIRED = "expired",
+}
```
derrikzzz: Should we be deriving the enum here from `medicationStatusEnum.enumValues` in `schema.ts` to prevent drift?

src/components/medication-stock/CreateModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+            />
+          ) : (
+            "An error has occurred while loading the dropdown options for brand + active ingredient. Please refresh the page."
+          )}
+
+          <RHFDropdown
+            name="stockStatus"
+            label="Status"
+            dropdownOptions={stockStatusDropdown}
+            className="mb-8"
```
derrikzzz, line 112: Should `stockStatus` be required here?

src/__tests__/pages/medication-stock/index.test.tsx
```diff
@@ -0,0 +1,379 @@
...
+            name: "Panadol",
+            activeIngredientName: "Paracetamol 500mg",
+          },
+        ],
+        isLoading: false,
+      },
+    );
+
+    mockTrpc.medicationStockRouter.create.useMutation.mockImplementation(
+      ({ onSuccess }) => {
```
derrikzzz: Am getting a typescript error here that `Binding element 'onSuccess' implicitly has an 'any' type` here!

src/components/medication-stock/CreateModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+                "flex-1 text-white px-4 py-1 rounded-lg font-medium",
+                form.formState.isSubmitting
+                  ? "bg-neutral-600"
+                  : "bg-green-600 hover:bg-green-700",
+              )}
+              disabled={form.formState.isSubmitting}
+              type="submit"
+            >
+              {form.formState.isSubmitting ? "Saving..." : "Save"}
+            </button>
```
derrikzzz: Would it be better if we use the `<Button>` component for consistency instead of raw `<button>` elements here?

src/components/medication-stock/SplittingModal.tsx
```diff
@@ -0,0 +1,122 @@
...
+        title="Add Split"
+        colour="indigo"
+        onClick={() => setSplits([...splits, stock])}
+        className="my-4"
+      />
+      <Button
+        title="Confirm"
+        colour="emerald"
+        onClick={() => console.log(splits)}
+        // onClick: validate parent qty === sum(all child qty) here, then bulk mutate
```
derrikzzz: nit: can comment here 
```suggestion
{/* TODO(separate PR): validate parent qty === sum(all child qty), then bulk mutate the splits */}
```

src/components/medication-stock/SplittingModal.tsx
```diff
@@ -0,0 +1,122 @@
...
+              value={split.location ? split.location : ""}
+              onChange={(e) =>
+                setSplits(
+                  splits.map((item, itemIndex) =>
+                    itemIndex === index
+                      ? { ...item, location: e.target.value }
+                      : item,
+                  ),
+                )
+              }
```
derrikzzz: Would it be better if we can replace these copies of the inline split pattern with a single updateSplit that takes in a index and path `updateSplit(index, patch)` helper method to reduce some level of duplication here

src/components/medication-stock/CreateModal.tsx
keiran0: Fixed [b6e297e](https://github.com/NUS-Project-SaBai/project-sabai/pull/130/commits/b6e297ef0ed6483b80b09d07e762911c0578c410)

src/lib/utils/medication-stock.ts
keiran0: Fixed [ff4a091](https://github.com/NUS-Project-SaBai/project-sabai/pull/130/commits/ff4a09152ec230c4f3e45ba7688cdf2d48022bff)

src/components/medication-stock/CreateModal.tsx
keiran0: Fixed in 3714a67 (commit message not reflective since I thought it was already committed though, excuse the error)

src/components/medication-stock/SplittingModal.tsx
keiran0: [631655e](https://github.com/NUS-Project-SaBai/project-sabai/pull/130/commits/631655e002c070634fadfcb1e6e28b5515d8549f) Fixed

src/__tests__/pages/medication-stock/index.test.tsx
keiran0: Fixed in 9a9b3ba

src/components/medication-stock/SplittingModal.tsx
keiran0: Good catch, fixed in 3b73c10

src/__tests__/pages/medication-stock/index.test.tsx
```diff
@@ -0,0 +1,387 @@
...
+  fireEvent,
+} from "@testing-library/react";
+import MedicationStockBasePage from "@/pages/medication-stock";
+import {
+  assertBreadcrumbs,
+  assertLoadingSpinner,
+  assertTableContents,
+} from "@/__tests__/utils/helper-functions";
+
+const mockStock = [
```
derrikzzz: nit: would it be better if we change `mockStock` to be `MOCK_STOCK` to match the fixture naming convention used in other places like `MOCK_VITALS` and `MOCK_VILLAGE_CODES` to name a few?

src/components/medication-stock/SplittingModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+    const newSplits = splits.map((item, itemIndex) =>
+      itemIndex === index
+        ? {
+            ...item,
+            ...patch,
+          }
+        : item,
+    );
+
+    setSplits(newSplits!);
```
derrikzzz: nit: was wondering if we need the guard here? The `Array.prototype.map` always return a new array, was thinking if the `!` can be dropped here so it can be:
```suggestion
    setSplits(newSplits);
```

src/components/medication-stock/EditingModal.tsx
```diff
@@ -0,0 +1,3 @@
+export default function EditingModal() {
+  return <h1>Editing</h1>;
```
derrikzzz: nit: would it be better to have comment here that it is a placeholder here temporarily? Then if we do grep sweep for TODO we can find some follow-up tasks that we can work on
```suggestion
// TODO(separate PR): implement stock editing form. Placeholder stub for now.
export default function EditingModal() {
  return <h1>Editing</h1>;
}
```

src/__tests__/pages/medication-stock/index.test.tsx
keiran0: Fixed in 099dab8

src/components/medication-stock/SplittingModal.tsx
keiran0: Fixed in 10233aa

src/components/medication-stock/EditingModal.tsx
keiran0: I've decided to remove `EditingModal`, to use a different style for editing instead (editablecell), removed and added back the conditionally rendered span element and added the comment that addresses your review similarly in ef45676

src/components/medication-stock/CreateModal.tsx
```diff
@@ -0,0 +1,136 @@
...
+              dropdownOptions={brandWithActiveIngredientData!.map((elem) => ({
+                value: elem.id.toString(),
+                label: `${elem.activeIngredientName} (${elem.name})`,
+              }))}
+              className="mb-4"
+              isRequired
+            />
+          ) : (
+            "An error has occurred while loading the dropdown options for brand + active ingredient. Please refresh the page."
+          )}
```
derrikzzz: Will it be better if we restructure to put the exception path first, and the dropdown is the happy path? 
```suggestion
{!brandWithActiveIngredientIsLoading &&
  (brandWithActiveIngredientIsError ? (
    "An error has occurred while loading the dropdown options for brand + active ingredient. Please refresh the page."
  ) : (
    <RHFDropdown
      name="medicationBrandId"
      label="Brand + Active Ingredient"
      dropdownOptions={brandWithActiveIngredientData!.map((elem) => ({
        value: elem.id.toString(),
        label: `${elem.activeIngredientName} (${elem.name})`,
      }))}
      className="mb-4"
      isRequired
    />
  ))}
```

src/lib/utils/medication-stock.ts
```diff
@@ -0,0 +1,23 @@
+import { MedicationStock } from "@/db/schema";
+
+import { medicationStatusValues } from "@/db/schema";
```
derrikzzz: Will it be better if this is merged as single import statement?
```suggestion
import { MedicationStock, medicationStatusValues } from "@/db/schema";
```

src/components/medication-stock/SplittingModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+            <td>Location:</td>
+            <td>{stock.location}</td>
+          </tr>
+          <tr>
+            <td>Active Ingredient</td>
+            <td>{stock.medicationActiveIngredientName}</td>
+          </tr>
+          <tr>
+            <td>Brand Id</td>
+            <td>{stock.medicationBrandName}</td>
```
derrikzzz: Should we be showing Brand Id or Brand Name here?

src/components/medication-stock/SplittingModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+        title="Add Split"
+        colour="indigo"
+        onClick={() => setSplits([...splits, stock])}
+        className="my-4"
+      />
+      <Button
+        title="Confirm"
+        colour="emerald"
+        onClick={() => console.log(splits)}
+        // TODO(separate PR): validate parent qty === sum(all child qty), then bulk mutate the splits
```
derrikzzz, line 114: Can we create a separate issue for this on the validation?

keiran0: This will be done together in the next phase for medication stock, validation built in

src/lib/utils/medication-stock.ts
keiran0: eb8d8f6 fixed

src/components/medication-stock/SplittingModal.tsx
keiran0: Fixed in 97c0f35

src/components/medication-stock/CreateModal.tsx
keiran0: I've moved the error block above the happy path so that it's a true happy path bace0ff

src/__tests__/pages/medication-stock/index.test.tsx
```diff
@@ -0,0 +1,387 @@
...
+    await user.click(brandIdButton!);
+
+    const brandDropdownButton = dialog.querySelector(
+      'button[name="medicationBrandId-1-dropdown-option"]',
+    );
+    await user.click(brandDropdownButton!);
+
+    await user.click(statusButton!);
+    const stockDropdownOption = dialog.querySelector(
+      'button[name="stockStatus-1-dropdown-option"]',
```
derrikzzz: Should this be `stockStatus-active-dropdown-option`, where does the 1 come from here in `'button[name="stockStatus-1-dropdown-option"]'`

Image

keiran0: Got it confused with the other dropdown field, fixed in 93ea233

derrikzzz: No worries! 👍

derrikzzz (approved):
Looks good to merge, hopefully this will unblock other issues haha ;D

PR #131, opened 2026-07-02
Medication schema changes

keiran0:
## Description

This PR:
- Adds 'remarks' column to all medication related tables
- Modifies the medication stock status enum, remove 'expired' because this is reflected in expiry date and shouldn't be captured as a state)

## Todo after merging in
- Remove 'reserved' state in the frontend stock state dropdown in future medication stock implementation (internal state only)


src/db/schema.ts
```diff
@@ -151,7 +156,8 @@ Medication Stock Table:
 - quantity: Quantity of the medication
 - expiry: Expiry date of the medication
 - location: Location of the medication
-- state: State of the medication (e.g., 'active', 'disposed', 'donated', 'expired')
+- state: State of the medication (e.g., 'active', 'disposed', 'donated', 'dispensed', 'reserved')
```
derrikzzz: Would it be better if we can update `disposed` to `discarded`, to be consistent with the enum here
```
export const medicationStatusEnum = pgEnum("medication_status", [
  "active",
  "discarded",
  "donated",
  "dispensed",
  "reserved",
]);
```

```suggestion
- state: State of the medication (e.g., 'active', 'discarded', 'donated', 'dispensed', 'reserved')
```

src/server/routers/medication_active_ingredient_router.ts
```diff
@@ -25,6 +26,7 @@ export const medicationActiveIngredientsRouter = router({
         name: zfd.text().nonoptional(),
         unitOfMeasurement: zfd.text().nonoptional(),
         fallBelow: zfd.numeric(z.number().int().positive()).nonoptional(),
+        remarks: zfd.text().nonoptional(),
```
derrikzzz: The remarks column is nullable in the DB schema but the procedure marks it as nonoptional  here — was this intentional? If not, should we change it to zfd.text() to match?

src/lib/utils/medication-stock.ts
```diff
@@ -13,6 +13,7 @@ export type CreateFormFields = {
   expiry: Date;
   location: string;
   stockStatus: StockStatus;
+  remarks: string;
```
derrikzzz: Should `remarks` be optional here?

src/server/routers/medication_active_ingredient_router.ts
keiran0: Fixed in 9f4c176

src/lib/utils/medication-stock.ts
keiran0: Fixed in f276c32

src/pages/medication-brands/index.tsx
```diff
@@ -68,6 +70,11 @@ function MedicationBrandsBasePage() {
                     {brand.name}
                   </span>
                 </TableCell>
+                <TableCell>
+                  <span className="text-sm font-medium text-slate-900">
+                    {brand.remarks}
+                  </span>
+                </TableCell>
```
derrikzzz, line 77: Not too sure about this and showing empty cells, but should we have fallback as part of UX improvement if there are no remarks to display (e.g using em dash, hyphen, showing N/A)?
```suggestion
<span className="text-sm font-medium text-slate-900">
  {brand.remarks || <span className="text-slate-400">—</span>}
</span>
```

derrikzzz: Created ticket for this in #173, think another PR on covering the fallback behavior for these will be suitable!

derrikzzz (approved):
Looks good to merge, have created issue https://github.com/NUS-Project-SaBai/project-sabai/issues/173 on covering the fallback behavior for empty cells for nullable fields!

PR #132, opened 2026-07-03
Consults schema and routers

derrikzzz:
## Description 

Adds data layer for doctor consultations: `consults` and `diagnosis` tables, tRPC routers for both, and a stub `/consults` page wired into the sidebar temporarily

### Schema (`src/db/schema.ts`)
- New `consults` table — one per visit (enforced by a unique constraint on `visit_id`), with `past_medical_history`, `consultation`, `treatment_plan`, `remarks`, a required `doctor_id` FK to `auth.users`, and a `visit_id` FK that cascades on visit deletion
- New `diagnosis` table 
- Added an `auth.users` schema stub so Drizzle can reference supabase-managed auth schema without migrating it

## Test plan
- [x] All 50 existing tests pass (`pnpm test:run`)
- [x] Typecheck and ESLint clean
- [x] Manual: run migration, create/update/delete a consult and diagnoses via the routers

### DIFF
--- a/src/db/schema.ts
+++ b/src/db/schema.ts
@@ -285,10 +285,12 @@ export const consults = pgTable("consults", {
   consultation: text("consultation"),
   treatmentPlan: text("treatment_plan"),
   remarks: text("remarks"),
-  doctorId: uuid("doctor_id").references(() => authUsers.id),
+  doctorId: uuid("doctor_id")
+    .notNull()
+    .references(() => authUsers.id),
   visitId: integer("visit_id")
     .notNull()
-    .references(() => visits.id, { onDelete: "cascade" }),
+    .references(() => visits.id),
 });
 
 export type Consult = typeof consults.$inferSelect;
@@ -307,7 +309,7 @@ export const diagnosis = pgTable("diagnosis", {
   category: varchar("category", { length: 255 }),
   consultId: integer("consult_id")
     .notNull()
-    .references(() => consults.id, { onDelete: "cascade" }),
+    .references(() => consults.id),
 });
 
 export type Diagnosis = typeof diagnosis.$inferSelect;
--- a/src/db/schema.ts
+++ b/src/db/schema.ts
@@ -1,5 +1,6 @@
 import {
   pgTable,
+  pgSchema,
   serial,
   varchar,
   boolean,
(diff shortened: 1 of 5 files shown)


src/db/schema.ts
```diff
@@ -259,5 +267,54 @@ export const eyesight = pgTable("eyesight", {
...
+  consultation: text("consultation"),
+  treatmentPlan: text("treatment_plan"),
+  remarks: text("remarks"),
+  doctorId: uuid("doctor_id")
+    .notNull()
+    .references(() => authUsers.id),
+  visitId: integer("visit_id")
+    .notNull()
+    .unique()
+    .references(() => visits.id, { onDelete: "cascade" }),
```
keiran0: Should the `onDelete` actions be 'restrict' instead? We don't have any app functionality to delete, but any attempts to delete visits should be blocked. 

I believe restrict is the default behaviour if unspecified by pgsql, would it be better if the reference options be left blank instead? 

Same for line 313.

src/db/schema.ts
```diff
@@ -259,5 +267,54 @@ export const eyesight = pgTable("eyesight", {
...
+  pastMedicalHistory: text("past_medical_history"),
+  consultation: text("consultation"),
+  treatmentPlan: text("treatment_plan"),
+  remarks: text("remarks"),
+  doctorId: uuid("doctor_id")
+    .notNull()
+    .references(() => authUsers.id),
+  visitId: integer("visit_id")
+    .notNull()
+    .unique()
```
keiran0: Should the unique constraint be removed here? The production database contains examples of many 1 (consult) to many (visits) relationship.

keiran0 (changes requested):
Some changes required for the schema. 

The page for consults appears to be a placeholder (due to my request to expedite consults schema into the codebase first), with some procedures for the diagnosis and consults routers defined but unused currently. Perhaps it would be good to review the routers and frontend properly when it's been actually implemented and the procedures used, so will approve these parts first and do a more thorough inspection when everything else is in order. (Alternative revert commits till f5d8dcb for this branch and move the other commits elsewhere, to be done together with the frontend)

src/db/schema.ts
derrikzzz: Ah okie I think we can follow that (option 2) to be unspecified(?), [95efc41](https://github.com/NUS-Project-SaBai/project-sabai/pull/132/commits/95efc41f8080868d2600266a04e40acecdb13813)

src/db/schema.ts
derrikzzz: remove unique constraint in [95efc41](https://github.com/NUS-Project-SaBai/project-sabai/pull/132/commits/95efc41f8080868d2600266a04e40acecdb13813), yeap I think we can follow that relationship

keiran0 (approved):
LGTM!

PR #143, opened 2026-07-06
pin pnpm version and update README accordingly

nathantew14:
pnpm now has version 11, which causes issues when attempting to install deps. pinned to v10.34.4 in package.json and update README accordingly (which was also missing instructions for installing pnpm). 

Also updated GitHub workflows to automatically use the pnpm version in package.json, instead of choosing a vague v10.

We should discuss in the future whether to migrate to pnpm 11 (the migration looks pretty trivial).


noahseethorcodes:
what's the impact of doing the migration? i.e. which things would we have to migrate?

PR #148, opened 2026-07-17
refactor: combine facial recognition endpoints

noahseethorcodes:
### What
Merges the two-step facial recognition search (`findFaceMatches` → `listMatchingPatients`) into a single `searchPatientsByPicture` endpoint that handles the full flow of Rekognition search plus patient DB lookup in one round trip. The frontend component (`MatchingPatients`) now calls one mutation and derives all UI states from its result.

### Why
The previous design exposed an unnecessary network round-trip and forced the client to shuttle raw AWS FaceMatch objects between two mutations, leaking Rekognition internals into the frontend. Combining them simplifies the data flow and removes a class of partial-failure states where the first call succeeds but the second fails silently.

### How
Backend: replaced `findFaceMatches` + `listMatchingPatients` with a single `searchPatientsByPicture` mutation on patientsRouter. The new endpoint calls searchFaceprint, extracts FaceIds, queries the patients table with inArray, and returns patient objects with image URLs. Still a mutation (not a query) because the base64 image payload can be large.

Frontend: rewired `MatchingPatients.tsx` to use the single mutation. Removed local state (`matchingPatients`), the intermediate `PatientWithImage` type, and the `react-hot-toast` error toasts in favour of an inline error state. The component now renders based on the mutation's `isIdle | isPending | isError | isSuccess` states directly.


src/server/routers/patients_router.ts
```diff
@@ -213,58 +213,38 @@ export const patientsRouter = router({
       return { success: !!result };
     }),
 
-  // find all matches
-  findFaceMatches: protectedProcedure
+  // searchPatientsByPicture uses AWS Rekognition to find matching patients based on a provided face image.
+  // It returns an array of patients whose rekognitionFaceId matches any of the FaceIds found in the search results.
+  // NOTE: Though this endpoint does not modify any data, it is a mutation because it receives a base64-encoded image as input, which can be large.
+  // Using a mutation allows for larger payloads compared to a query.
+  searchPatientsByPicture: protectedProcedure
```
noahseethorcodes, line 246: tRPC's query procedures are transported over HTTP GET requests, which encode all input parameters into the URL as query strings. A base64-encoded webcam image is typically 100–500KB in size, far exceeding the practical URL length limits enforced by browsers, proxies, and the Node.js HTTP server that Next.js runs on. Sending an image through a GET URL would reliably cause a 414 (URI Too Long) error before the request even reaches the handler.

tRPC's mutation procedures, by contrast, are transported over HTTP POST requests, which carry their payload in the request body. The body has no meaningful size limit imposed by the protocol itself — it is governed only by server configuration, which defaults to several megabytes. This makes POST the only viable transport for large inputs like images.

src/components/scanFace/MatchingPatients.tsx
```diff
@@ -1,71 +1,58 @@
...
-type PatientWithImage = Patient & {
-  patientImageUrl: string | null;
-};
-
 export default function MatchingPatients({
   imgDetails,
   setMode,
 }: {
-  imgDetails: string | null;
+  imgDetails: string;
```
noahseethorcodes, line 15: `imgDetails` prop narrowed from `string | null` to `string`. The parent component already guards against null via `&& imgDetails` before rendering this component, so this makes the contract explicit and removes the need for the ! assertion.

keiran0 (changes requested):
LGTM, thanks for catching my coding brain farts! However, the tests still need to be updated (`src/__tests__/pages/scan-face.test.tsx`). Alternatively, the tests expected to fail after the change can be deleted and rewritten in another PR.

src/__tests__/pages/scan-face.test.tsx
```diff
@@ -26,10 +25,7 @@ vi.mock("@/utils/trpc", () => ({
...
         useMutation: vi.fn(),
       },
-      findFaceMatches: {
-        useMutation: vi.fn(),
-      },
-      listMatchingPatients: {
+      searchPatientsByPicture: {
         useMutation: vi.fn(),
       },
     },
```
noahseethorcodes, line 31: updated tests.
summary: 
- replaced `findFaceMatches` + `listMatchingPatients` mocks with single `searchPatientsByPicture` mock
- removed the second loading spinner test (only one loading state now)

src/__tests__/pages/scan-face.test.tsx
```diff
@@ -60,12 +56,15 @@ describe("ScanFacePage", () => {
...
-    });
+    mockTrpc.patientsRouter.searchPatientsByPicture.useMutation.mockReturnValue(
+      {
+        mutate: vi.fn(),
+        isIdle: false,
+        isPending: false,
+        isError: false,
+        data: [],
+      },
+    );
```
noahseethorcodes, line 67: by default, the mock will return a completed state (i.e. no pending idle or error) with empty data

keiran0 (approved):
LGTM, thanks for working on this and catching the unnecessary round trip!

PR #150, opened 2026-07-19
fix(seed): assign a visit to every patient in visits seed

derrikzzz:
Closes #149 

## Description
- Updates visits seed (`supabase/seeds/05_visits.sql`), so that every patient in the seed script will be assigned at least 1 visit

## Why
- Previously the seed only created visits for patients 1-8, leaving patients 9-15 visitless. In Supabase, there is no patient without a visit.

## Test plan
- [ ] Confirmed `01_patients.sql` seeds 15 patients (IDs 1–15).
- [ ] Confirmed every patient ID 1–15 is now referenced in `05_visits.sql`.
- [ ] Reset and re-seed the database, then verify no patient is without a visit (expect 0 rows), can execute the query in pgAdmin / Tableplus / other DB GUIs if applicable
- [ ] Load the Vision (and Vitals) page and confirm all 15 patients render, ordered by latest visit date with none in the visit-less bucket.
```
SELECT p.id
FROM patients p
LEFT JOIN visits v ON v.patient_id = p.id
WHERE v.id IS NULL;
```


keiran0 (approved):
LGTM!

PR #154, opened 2026-07-21
Fix unintended code change when merging branches

noahseethorcodes:
Changed a `useEffect` function to use an old query while resolving merge conflicts. Need to change it to use the currently existing query if not the build will fail


keiran0 (approved):
LGTM!

PR #159, opened 2026-07-24
158 eyesight visual acuity validation

### LINKED ISSUE #158: Adding of validation for eyesight visual acuity fields (Snellen notation)
## Description

- The eyesight record fields (visual acuity, pinhole, astigmatism, prescribed glasses) currently accept **any free-text string** at every layer of the stack. There is no validation that a value like `6/6`, `6/12`, `6/60` is well-formed, so invalid entries (`66`, `6-6`, `garbage`, empty-but-present strings) are saved silently. We would like to prevent this and introduce validation at the router level such that

## Current behavior

No validation exists at any of the three layers:

| Layer | Location | Current state |
|-------|----------|---------------|
| Database | `src/db/schema.ts` (`eyesight` table) | All eye fields are plain `text()` columns — no constraint |
| API (tRPC) | `src/server/routers/eyesight_router.ts:28-39` | Every field typed as `z.string().optional()` — accepts any string |
| Form | `src/components/vision/EyesightForm.tsx:155-161` | Generic `<RHFInput type="text">`, no `pattern` or RHF validation rules |

## Expected behavior

- Visual acuity fields should validate against accepted clinical notation and reject malformed input with error message, both on client and server(?)

## Acceptance criteria
- [ ] Zod schema in `eyesight_router.ts` validates acuity fields per agreed rules
- [ ] Client-side form validation surfaces a clear inline error before submit
- [ ] Invalid values are rejected by the API even if the client is bypassed
- [ ] Tests cover valid, invalid, and empty/optional cases

derrikzzz:
Closes #158 

## Description

- Adds format validation for eyesight visual acuity and pinhole fields which were previously stored as text-free unvalidated. Validation is defined and enforced identically on client (form) and server (tRPC). Notation should follow the clinic's VA measurement SOP from last year.

## Changes

- `src/lib/validation/eyesight.ts` — shared validation primitives (safe to import on both client and server):
- `src/server/routers/eyesight_router.ts` — `createEyesightInput/updateEyesightInput` now validate the acuity + pinhole fields via the shared schemas. Normalization happens here on parse(), so the DB stores canonical values.
-` src/components/vision/EyesightForm.tsx` — the Visual Acuity and Pinhole inputs validate against the same schemas via react-hook-form, so **invalid entries are caught before submit** with the same rule the server enforces.
- src/__tests__/lib/validation/eyesight.test.ts — addition of unit tests.

## Automated Test Plan
- [ ] pnpm test:run src/__tests__/lib/validation/eyesight.test.ts — test case passes
- [ ] npx tsc --noEmit and eslint clean on the changed files

## Manual Test Plan
- [x] Open a patient's vision/update-glasses page.
- [ ] Enter an invalid acuity (e.g. 20/20 or 6/7) in a Degree field → inline error shown, save blocked.
- [ ] Enter 6/9 +2 and CF → accepted.
- [ ] Enter NI in a Pinhole field → accepted; enter NI in a Degree field → rejected.
- [ ] Enter lowercase lp, save, reload → persists as LP.
- [ ] Leave a field blank → saves fine (optional).
- [ ] Confirm astigmatism / prescribed-glasses fields still accept free text.

### DIFF
--- a/src/lib/validation/eyesight.ts
+++ b/src/lib/validation/eyesight.ts
@@ -22,7 +22,13 @@ export const PINHOLE_REGEX = new RegExp(`^(${ACUITY_PATTERN}|NI)$`, "i");
 
 // Optional, trimmed string field; empty string means "not recorded".
 const optionalField = (regex: RegExp, message: string) =>
-  z.string().trim().regex(regex, message).or(z.literal("")).optional();
+  z
+    .string()
+    .trim()
+    .regex(regex, message)
+    .toUpperCase()
+    .or(z.literal(""))
+    .optional();
 
 export const visualAcuitySchema = optionalField(
   VISUAL_ACUITY_REGEX,
--- a/src/lib/validation/eyesight.ts
+++ b/src/lib/validation/eyesight.ts
@@ -4,10 +4,15 @@ import { z } from "zod";
  * Shared, dependency-free validation primitives for eyesight fields,
  * safe to import on both server and client. Request-shape schemas live in the
  * router.
+ * 
+ * Per clinic SOP: Snellen fractions 6/6..6/120 (6m tumbling-E chart) with an
+ * optional "+N" for N letters read on the next line (e.g. "6/9 +2"), plus the
+ * low-vision codes CF (counting fingers), HM (hand movement),
+ * LP (light perception), NLP (no light perception).
  */
 
-// Snellen fractions (6/6..6/60, optional +/- suffix) plus low-vision codes: CF (counting fingers), HM (hand motion), PL/NPL (perception of light).
-const ACUITY_PATTERN = "6\\/(6|9|12|18|24|36|60)[+-]?|CF|HM|PL|NPL";
+const ACUITY_PATTERN =
+  "6\\/(6|9|12|18|24|36|60|120)(\\s*\\+\\d{1,2})?|CF|HM|LP|NLP";
 
 /** Visual acuity (degree) fields. Case-insensitive. */
 export const VISUAL_ACUITY_REGEX = new RegExp(`^(${ACUITY_PATTERN})$`, "i");
@@ -21,10 +26,10 @@ const optionalField = (regex: RegExp, message: string) =>
 
 export const visualAcuitySchema = optionalField(
   VISUAL_ACUITY_REGEX,
-  "Must be Snellen notation (e.g. 6/6, 6/12) or CF, HM, PL, NPL",
+  "Must be VA notation (e.g. 6/9, 6/9 +2, 6/120) or CF, HM, LP, NLP",
 );
 
 export const pinholeSchema = optionalField(
   PINHOLE_REGEX,
-  "Must be Snellen notation (e.g. 6/6, 6/12), CF, HM, PL, NPL, or NI",
+  "Must be VA notation (e.g. 6/9, 6/9 +2, 6/120), CF, HM, LP, NLP, or NI",
 );
--- a/src/lib/validation/eyesight.ts
+++ b/src/lib/validation/eyesight.ts
@@ -1,0 +1,30 @@
+import { z } from "zod";
+
+/**
+ * Shared, dependency-free validation primitives for eyesight fields,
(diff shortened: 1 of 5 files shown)


src/lib/validation/eyesight.ts
```diff
@@ -0,0 +1,41 @@
...
+ * safe to import on both server and client. Request-shape schemas live in the
+ * router.
+ * 
+ * Per clinic SOP: Snellen fractions 6/6..6/120 (6m tumbling-E chart) with an
+ * optional "+N" for N letters read on the next line (e.g. "6/9 +2"), plus the
+ * low-vision codes CF (counting fingers), HM (hand movement),
+ * LP (light perception), NLP (no light perception).
+ */
+
+const ACUITY_PATTERN =
```
keiran0, line 14: Should "6/15" and "6/7.5" also be captured in the acuity pattern? These values exist in the database.

keiran0 (changes requested):
Requesting changes for two extra visual acuity values. Although not in the clinic SOP, there are values for something like `6/9-2` which might be something we need to clarify with the clinic heads before we proceed with this PR. I've recorded the query output, to ask during a further meeting. This PR might be shelved until we get the greenlight

src/lib/validation/eyesight.ts
derrikzzz: Added 6/7.5 and 6/15 to the `ACUITY_PATTERN` here [f41321a](https://github.com/NUS-Project-SaBai/project-sabai/pull/159/commits/f41321aa67554bd16fb44bf2369cc02f58896d56)

src/lib/validation/eyesight.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Per clinic SOP: Snellen fractions 6/6..6/120 (6m tumbling-E chart) with an
+ * optional "+N"/"-N" suffix for letters read over/under on the next line
+ * (e.g. "6/9 +2", "6/9-2"), plus the low-vision codes CF (counting fingers),
+ * HM (hand movement), LP (light perception), NLP (no light perception).
+ *
+ * 6/7.5 and 6/15 are also accepted — they appear in the old db.
+ */
+
+const ACUITY_PATTERN =
+  "6\\/(7\\.5|6|9|12|15|18|24|36|60|120)(\\s*[+-]\\d{1,2})?|CF|HM|LP|NLP";
```
keiran0: Would it be better to disallow whitespaces in between altogether? It might be easier for users to remember not to put whitespace in, instead of remembering that whitespace before the + is acceptable while after the + is unacceptable.

src/lib/validation/eyesight.ts
```diff
@@ -0,0 +1,43 @@
...
+    .string()
+    .trim()
+    .regex(regex, message)
+    .toUpperCase()
+    .or(z.literal(""))
+    .optional();
+
+export const visualAcuitySchema = optionalField(
+  VISUAL_ACUITY_REGEX,
+  "Must be VA notation (e.g. 6/9, 6/9 +2, 6/9-2, 6/120) or CF, HM, LP, NLP",
```
keiran0: Reminder to also edit the messages here if the line 17 comment is also addressed!

keiran0 (changes requested):
Some changes with the consistency of spaces, otherwise LGTM!

src/lib/validation/eyesight.ts
derrikzzz: Disallow whitespace here [7e782bc](https://github.com/NUS-Project-SaBai/project-sabai/pull/159/commits/7e782bc1725c9c9fd663ddbf4ccffb5a0c78d8bd), will flag on blur

src/lib/validation/eyesight.ts
derrikzzz: Edited message in same commit here too [7e782bc](https://github.com/NUS-Project-SaBai/project-sabai/pull/159/commits/7e782bc1725c9c9fd663ddbf4ccffb5a0c78d8bd)

keiran0 (approved):
LGTM!

PR #161, opened 2026-07-27
Stock Changes and Orders Table

keiran0:
## Description 

This PR creates 2 new tables -  the `stock_changes` and `orders`. The PR is a brief description on the changes, but more details including implementation, examples, context, and other considerations on how this table will work in conjunction with the `orders` table is in the design document on Google Drive.

## Changes to the `orders` table (compared to old codebase)

There are major changes that the new table seeks to address in the old codebase.
- `order` has been changed to `orders` to stick with grammatical conventions (as well as avoiding keyword clash)
- `medication_review` and `order` has been merged. Previously, we found out that there are medication reviews without any corresponding order. Those were due to medication reviews being used as a standalone feature for stocktaking/admin (to tally the medicines on their end during stocktake). However, with the pharm management overhaul, stocktaking is expected to be easier and there will not be any need for such workarounds (ideally). As there should be a 1-1 relationship between orders and reviews, the tables have been combined. 
- `order.remarks` has been removed as the column contains nothing.
- `order.notes` has been renamed to `orders.dosage_instructions` as that is what the frontend displays, but backend saves it as `notes` which is a bit confusing.

## New aspects of `orders` table

`stock_change_id` refers to the `stock_change.id` field that has the stock flipped from 'active' to 'reserved'. As a brief explanation, the 'reserved' state is an internal only state to prevent race conditions during the period of doctor order to patient receiving medications. Medication in this state will not be editable (other than location, though unlikely) until approved/rejected (changed to dispensed/active state)

## Creation of the `stock_changes` table

The `stock_changes` table serves to provide an auditable, append-only trail of who changed what wrt to the medication stock. This can refer to a stock's location, remarks (blocked by PR #131) or status. The stock change operations include doctor ordering medication during a consult, manual splitting of stock, approval/rejection of an order, editing of stock fields, and adding stock. All changes are captured in the `stock_changes` table, including the doctor who created the order, the PL team member who approved/rejected the order, anyone who edits details about the stock, or adds a stock.

## Todo before undrafting (resolved):
- Permission is being sought to exclude medication reviews without orders in the future database migration to make sure everything has been accounted for and no important data is being dropped. -> not needed, we're not dropping the data since the previous database is being kept. However, moving forward, we're actually preserving the purpose of the changes in the stock by making use of the stock splitting/state features. -> (resolved 1 Aug) PL head has given greenlight to drop that column for future DB. The new stock statuses should account for any rebalancing for stocktaking.

### DIFF
--- a/src/db/schema.ts
+++ b/src/db/schema.ts
@@ -376,8 +376,8 @@ export const stockChanges = pgTable("stock_changes", {
     .notNull()
     .references(() => medicationStock.id),
   field: varchar("field").notNull(),
-  from: varchar("from").notNull(),
-  to: varchar("to").notNull(),
+  previousValue: varchar("previous_value").notNull(),
+  newValue: varchar("new_value").notNull(),
   userId: uuid("user_id")
     .notNull()
     .references(() => authUsers.id),
--- a/src/db/schema.ts
+++ b/src/db/schema.ts
@@ -383,3 +383,23 @@ export const stockChanges = pgTable("stock_changes", {
     .references(() => authUsers.id),
   createdAt: timestamp("created_at").defaultNow().notNull(),
 });
+
+/*
+Orders Table:
+- id: Primary key, auto-incrementing integer.
+- consultId: Foreign key referencing the consult in which the order was created.
+- dosageInstructions: The dosage instructions for the order.
+- status: The order status. Can be 'PENDING', 'CANCELLED', or 'APPROVED'.
+- stockChangeId: Foreign key representing the entry in the stockChange table that displays stocks going from 'active' to 'reserved' state.
+*/
+export const orders = pgTable("orders", {
+  id: serial("id").primaryKey(),
+  consultId: integer("consult_id")
+    .notNull()
+    .references(() => consults.id),
+  dosageInstructions: text("dosage_instructions").notNull(),
+  status: orderStatusEnum("status").notNull().default("PENDING"),
+  stockChangeId: integer("stock_change_id")
+    .notNull()
+    .references(() => stockChanges.id),
+});
--- a/src/db/schema.ts
+++ b/src/db/schema.ts
@@ -31,6 +31,11 @@ export const orderStatusEnum = pgEnum("order_status", [
   "CANCELLED",
   "PENDING",
 ]);
+export const stockChangeFieldEnum = pgEnum("stock_change_field_enum", [
+  "location",
+  "quantity",
+  "stock_status",
+]);
 
 // Define tables
 
@@ -375,7 +380,7 @@ export const stockChanges = pgTable("stock_changes", {
   stockId: integer("stock_id")
     .notNull()
     .references(() => medicationStock.id),
-  field: varchar("field").notNull(),
+  field: stockChangeFieldEnum("field").notNull(),
   previousValue: varchar("previous_value").notNull(),
   newValue: varchar("new_value").notNull(),
   userId: uuid("user_id")
--- a/src/db/schema.ts
+++ b/src/db/schema.ts
@@ -365,8 +365,8 @@ Stock Changes Table:
 - id: Primary key, auto-incrementing integer.
 - stockId: Foreign key referencing the stock that was changed.
 - field: The field of the row that was changed in the medication_stock table.
-- from: The previous value of the changed field.
-- to: The new value of the changed field.
+- previousValue: The previous value of the changed field.
+- newValue: The new value of the changed field.
 - userId: Foreign key referencing the user who changed the field.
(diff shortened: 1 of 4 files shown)


src/db/schema.ts
```diff
@@ -355,5 +360,46 @@ export const diagnosis = pgTable("diagnosis", {
...
+  id: serial("id").primaryKey(),
+  consultId: integer("consult_id")
+    .notNull()
+    .references(() => consults.id),
+  dosageInstructions: text("dosage_instructions").notNull(),
+  status: orderStatusEnum("status").notNull().default("PENDING"),
+  stockChangeId: integer("stock_change_id")
+    .notNull()
+    .references(() => stockChanges.id),
+});
```
derrikzzz, line 458: Will it be better if we can include type exports here such as over here, rationale is on consistency here in `schema.ts`
```suggestion
export type StockChange = typeof stockChanges.$inferSelect;
export type NewStockChange = typeof stockChanges.$inferInsert;
export type Order = typeof orders.$inferSelect;
export type NewOrder = typeof orders.$inferInsert;
```

src/db/schema.ts
```diff
@@ -355,5 +360,46 @@ export const diagnosis = pgTable("diagnosis", {
...
+- userId: Foreign key referencing the user who changed the field.
+- createdAt: The timestamp of when the field was changed.
+*/
+export const stockChanges = pgTable("stock_changes", {
+  id: serial("id").primaryKey(),
+  stockId: integer("stock_id")
+    .notNull()
+    .references(() => medicationStock.id),
+  field: varchar("field").notNull(),
+  from: varchar("from").notNull(),
```
derrikzzz: nit: can we use `previousValue` and `newValue` here? then the DB columns for this will be `previous_value` and `new_value`

src/db/schema.ts
```diff
@@ -355,5 +360,46 @@ export const diagnosis = pgTable("diagnosis", {
...
+- to: The new value of the changed field.
+- userId: Foreign key referencing the user who changed the field.
+- createdAt: The timestamp of when the field was changed.
+*/
+export const stockChanges = pgTable("stock_changes", {
+  id: serial("id").primaryKey(),
+  stockId: integer("stock_id")
+    .notNull()
+    .references(() => medicationStock.id),
+  field: varchar("field").notNull(),
```
derrikzzz: Would it be better if we have a `pgEnum` for this? I'm thinking this can only take on a few values like `location`, `quantity` etc.

src/db/schema.ts
```diff
@@ -26,6 +26,11 @@ export const medicationStatusEnum = pgEnum("medication_status", [
   "donated",
   "expired",
 ]);
+export const orderStatusEnum = pgEnum("order_status", [
+  "APPROVED",
+  "CANCELLED",
+  "PENDING",
+]);
```
derrikzzz, line 45: Would it be better if we standardise whether to use upper case or lower case here?

src/db/schema.ts
```diff
@@ -355,5 +360,46 @@ export const diagnosis = pgTable("diagnosis", {
...
+/*
+Orders Table:
+- id: Primary key, auto-incrementing integer.
+- consultId: Foreign key referencing the consult in which the order was created.
+- dosageInstructions: The dosage instructions for the order.
+- status: The order status. Can be 'PENDING', 'CANCELLED', or 'APPROVED'.
+- stockChangeId: Foreign key representing the entry in the stockChange table that displays stocks going from 'active' to 'reserved' state.
+*/
+export const orders = pgTable("orders", {
+  id: serial("id").primaryKey(),
```
derrikzzz, line 449: Would it be better if we can have a `createdAt` timestamp here?

src/db/schema.ts
keiran0: Added in 7d57d05

src/db/schema.ts
keiran0: Added in 5cd29e7

src/db/schema.ts
keiran0: I was thinking of evaluating it again in the future since the current uppercase is to keep in line with the previous DB. There are also other constants that follow different conventions (camelcase for the referral statuses etc) and we can discuss standardising it in the future. #172

src/db/schema.ts
keiran0: Captured by `stockChanges` table as a transaction with orders (resolving as further discussed in group)

src/db/schema.ts
keiran0: Added in 91f1d15

derrikzzz (approved):
Looks good to merge in! 🚀

PR #166, opened 2026-08-01
163 add referral routers

### LINKED ISSUE #163: Add referral routers
## Description

Referrals are created during a consult when the doctor decides a patient needs to be seen elsewhere or by a specialist. The referral lifecycle is tracked separately from the consult on a Referral Board, where doctors and medical studentscan update state and record outcomes.

| Procedure | Type | Description |
|--|--|--|
| `referralRouter.create` | mutation | Create a referral after a consult is saved. Accepts `consultId`, `referredFor`, `referralNotes`. State defaults to `"New"`. |
| `referralRouter.list` | query | List all referrals enriched with patient name, doctor ID, and consult date via `Referral → Consult → Visit → Patient` joins. Ordered by most recent. |
| `referralRouter.getById` | query | Fetch a single enriched referral by ID. |
| `referralRouter.getByConsultId` | query | Fetch all referrals for a given consult. |
| `referralRouter.updateState` | mutation | Update `referralState` and `referralOutcome` only. |
| `referralRouter.delete` | mutation | Delete a referral by ID. |

## Relevant Data model

Referrals table already exists in schema (`referredFor`, `referralNotes`, `referralState`, `referralOutcome`, `consultId`, `createdAt`). No migrations needed.

`referralState` enum values: `New`, `Seen`, `Outgoing`, `Completed`, `CompletedSuccess`, `CompletedFailure`, `None`.

## Additional Notes
- `referred_for` and `referral_notes` live only on the Referrals table, not on Consult, to avoid duplication.
- State defaults to `"New"` on creation, not null.
- All procedures use `protectedProcedure`.

## Acceptance criteria
- [ ] All procedures respond correctly
- [ ] `updateState` is able to reject invalid enum values (not in the `referralStateSchema`)
- [ ] Type checks pass

derrikzzz:
Closes #163 

## Description 

- Adds a tRPC referral router (`referralRouter`) exposing CRUD operations over the `referrals` table, wired into the app router. Includes the repo's first router-level integration tests too ;D

## Changes

- `src/server/routers/referral_router.ts` — new router with six procedures:
  - `list` — all referrals, enriched with patient/doctor/consult data
  - `getById` — single enriched referral, or `null` if not found
  - `getByConsultId` — all referrals for a consult
  - `create` — insert a referral (defaults `referralState` to `New` )
  - `updateState` — update state + outcome
  - `delete` — remove a referral, returns `{ success }`
- `src/test-setup.ts` — load `.env` via `dotenv/config` and guard the DOM  shims with a `typeof window` check so node-environment tests can run alongside the existing jsdom component tests.
- `src/__tests__/server/routers/referral_router.test.ts` — adding of integration tests to procedures

## Test plan
- Prerequisites: local Supabase running with migrations + a seeded user.
- Run the referral tests:

```
npx vitest run src/__tests__/server/routers/referral_router.test.ts --no
```

- [ ] `create()` defaults state to `New` and notes to `null` when omitted
- [ ] `list()` returns rows enriched via the patient/consult joins (`patientName`, `doctorId`)
- [ ] `getById()` returns the enriched referral, and `null` for a missing id
- [ ] `getByConsultId()` returns every referral for a consult (referral have FK to consult)
- [ ] `updateState()` rejects a state outside the enum (zod guard / validation)
- [ ] `delete()` reports `success: true` on a real delete, `false` on a repeat

### DIFF
--- a/src/__tests__/server/routers/referral_router.test.ts
+++ b/src/__tests__/server/routers/referral_router.test.ts
@@ -1,0 +1,214 @@
+// @vitest-environment node
+//
+// Integration tests for the referral tRPC router. These run against a real
+// Postgres (the local Supabase instance on 127.0.0.1:54322) so the joins,
+// enum handling, and FK constraints are exercised for real
+// Prereqs: `supabase start` (or `pnpm seed:db`) so migrations are applied, and at least one auth.users row (`pnpm seed:users`).
+//
+// Runs in the node environment (not jsdom) because src/lib/envVariables.ts
+// switches on `typeof window` and only exposes DATABASE_URL server-side.
+import { describe, it, expect, beforeAll, afterAll, afterEach } from "vitest";
+import { eq } from "drizzle-orm";
+import { appRouter } from "@/server/routers/_app";
+import { db } from "@/db/drizzle";
+import type { Context } from "@/server/context";
+import {
+  referrals,
+  consults,
+  visits,
+  patients,
+  villageCodes,
+  authUsers,
+} from "@/db/schema";
+
+// Fixture inserted into Postgres for the FK chain (not mocks).
+const SEED_PATIENT = {
+  name: "Referral Test Patient",
+  gender: "female",
+  drugAllergy: "NKDA", // NKDA stands for "No Known Drug Allergies"
+  dateOfBirth: new Date("1990-01-01"),
+  hasPoorCard: false,
+  hasBS2Card: false,
+  hasSabaiCard: false,
+  patientImagePublicId: "test/placeholder",
+} as const;
+
+const VILLAGE_CODE = {
+  code: "SATBB",
+  name: "SATBB Village",
+  colorHex: "#C0392B",
+};
+
+// protectedProcedure only reads ctx.user, so a minimal context is enough.
+function callerFor(userId: string) {
+  return appRouter.createCaller({ user: { id: userId } } as unknown as Context);
+}
+
+describe("referralRouter (integration)", () => {
+  let doctorId: string;
+  let villageCodeId: number;
+  let patientId: number;
+  let visitId: number;
+  let consultId: number;
+  let caller: ReturnType<typeof callerFor>;
+
+  beforeAll(async () => {
+    // A consult FK-references auth.users; reuse an existing seeded user.
+    const [doctor] = await db
+      .select({ id: authUsers.id })
+      .from(authUsers)
+      .limit(1);
+    if (!doctor) {
+      throw new Error(
+        "No auth.users row found. Run `pnpm seed:all` against local Supabase " +
+          "before running the router integration tests.",
+      );
+    }
+    doctorId = doctor.id;
+
+    [{ id: villageCodeId }] = await db
+      .insert(villageCodes)
+      .values({
+        code: VILLAGE_CODE.code,
+        name: VILLAGE_CODE.name,
+        colorHex: VILLAGE_CODE.colorHex,
+      })
+      .returning({ id: villageCodes.id });
+
+    [{ id: patientId }] = await db
+      .insert(patients)
+      .values(SEED_PATIENT)
+      .returning({ id: patients.id });
+
+    [{ id: visitId }] = await db
+      .insert(visits)
+      .values({ patientId, villageCodeId })
+      .returning({ id: visits.id });
+
+    [{ id: consultId }] = await db
+      .insert(consults)
+      .values({ doctorId, visitId })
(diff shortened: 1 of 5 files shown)


src/__tests__/server/routers/referral_router.test.ts
```diff
@@ -0,0 +1,214 @@
...
+  let consultId: number;
+  let caller: ReturnType<typeof callerFor>;
+
+  beforeAll(async () => {
+    // A consult FK-references auth.users; reuse an existing seeded user.
+    const [doctor] = await db
+      .select({ id: authUsers.id })
+      .from(authUsers)
+      .limit(1);
+    if (!doctor) {
```
keiran0: Should we clarify the naming here? Since it's not just doctors, it might be better to name this variable 'user' instead.

src/test-setup.ts
```diff
@@ -1,8 +1,15 @@
 import "@testing-library/jest-dom";
+// Load .env into process.env so server/router integration tests (node
+// environment) can read DATABASE_URL and the other server-only vars validated
+// by src/lib/envVariables.ts. Harmless for jsdom component tests.
+import "dotenv/config";
```
keiran0: Since this uses `.env`, should there be a guard to ensure that the database in the `.env` is actually localhost and not supabase? A point of concern here is that if someone has the development/production server's credentials on the `.env` and runs tests (though unlikely, we should guard against it), it could potentially wipe real data.

src/__tests__/server/routers/referral_router.test.ts
```diff
@@ -0,0 +1,214 @@
+// @vitest-environment node
+//
+// Integration tests for the referral tRPC router. These run against a real
+// Postgres (the local Supabase instance on 127.0.0.1:54322) so the joins,
```
keiran0: Seems this comment is an assumption, see comment on `test-setup.ts`. Would it be better to clarify that it uses the postgres tied to the `.env` instead of explicitly stating that it runs on localhost 54322?

keiran0 (changes requested):
Just some preliminary feedback first, will verify again with the test environment setup. Thanks for setting this up, it should help us with more tests going forward!

src/__tests__/server/routers/referral_router.test.ts
derrikzzz: Renamed here [bcfc86b](https://github.com/NUS-Project-SaBai/project-sabai/pull/166/commits/bcfc86bd8ad3379415f9250ef6df251ab8e141b1)

src/test-setup.ts
derrikzzz: Added `isDatabaseLocal()` method here in `environmentUtils.ts` [c5eb83d](https://github.com/NUS-Project-SaBai/project-sabai/pull/166/commits/c5eb83da455ee9bd0520be9123adfca704fa32f0)

I believe this can be the file where other helpers to answer a boolean question about the current environment can live in

src/__tests__/server/routers/referral_router.test.ts
derrikzzz: Clarified here bcfc86bd8ad337

src/test-setup.ts
```diff
@@ -1,8 +1,26 @@
...
+import { isDatabaseLocal } from "@/lib/environmentUtils";
+
+// Guard: refuse to run node-environment (server/integration) tests against a
+// non-local database. This prevents accidental data loss if someone has
+// dev/production credentials in their .env and runs the test suite.
+if (typeof window === "undefined" && !isDatabaseLocal()) {
```
keiran0: Should this be an 'or' condition instead? This guard clause wouldn't catch anything if the window was defined and the database wasn't local.

keiran0 (changes requested):
Just a small change to the guard

src/test-setup.ts
keiran0: Would it be better to just reject if the database isn't local? If typeof window is undefined it returns immediately (line 24).

```
if (!isDatabaseLocal()) {
...
}
```

keiran0 (changes requested):
A change to the guard clause

src/test-setup.ts
derrikzzz: In this case, will this not block running of tests that have nothing to do with db?

Or another approach is that I can update the method to return true when `DATABASE_URL` is absent?

keiran0: Good point

keiran0 (approved):
LGTM!

PR #174, opened 2026-08-02
171 edit patient details from vitals page

### LINKED ISSUE #171: Allow editing patient details from the vitals page
## Description

- Doctors and medical students can catch stale or incorrect villager / children data (mispelled name, updated drug allergy). The current approach on viewing the patient details is by appending the patient id as the slug, but that would not allow us to edit.

## Current state
- Backend already supports edits — `patients_router.update` exists (`src/server/routers/patients_router.ts:196`).
- `/patient/[id]` displays patient details but is **read-only** 
- `/vitals/[id]` only edits the per-visit vitals record, maybe on that page we should have a "Edit patient details" button
- Add patient-detail editing, **surfaced from the vitals page via a modal/link**, reusing the existing `patients_router.update` mutation. DO **NOT** embed demographic fields inline in the vitals `<form>` — that form's `Save` submits a visit-scoped vitals payload, and mixing patient-level fields into it muddies what "Save" means.

## Acceptance Criteria
- [ ] Patient details can be edited from `/patient/[id]`
- [ ] Patient details can be edited from `/vitals/[id]` via a modal, without touching the vitals form ("Edit Patient Details" button to open that modal)
- [ ] Edits persist via `patients_router.update` and the UI reflects updated values immediately
- [ ] Vitals save and patient-detail save remain independent actions

derrikzzz:
Closes #171 

## Description

Adds the ability to edit an existing patient's details (including re-capturing their photo) from both the patient profile page and the vitals page. Previously patient records were create-only; there was no way to fix bad typo or replace bad photo after registration of patients

### 1. Edit Form component
- New `EditPatientForm` component, prefilled from the patient and persisted via patientsRouter.update. Rendered in a modal from both `patient/[id]` and `vitals/[id]`.
- Two-column field layout; modal widened to `max-w-2xl` via a new optional `maxWidthClassName` prop on Modal (default unchanged for every other modal used).
- Date of birth is normalized to yyyy-MM-dd (toDateInputValue) so the native date input prefills instead of rendering blank

### 2. Photo recapture
- WebcamInput wired into the edit form. The new image is only sent when the user actually retakes it; otherwise existing photo is untouched.
- On the backend, update now uploads the new image and regenerates the Rekognition faceprint in parallel via `Promise.all`, so face-search stays in sync with the new photo.
- Old Cloudinary image and Rekognition faceprint are cleaned up (best-effort delivery). New helpers: `deleteFromCloudinary`, `deleteFaceprint`.

### 3. Duplicate ID protection
- update procedure rejects an id that already belongs to a different patient (CONFLICT), and the form surfaces that message in a toast.

## Test plan
Manual — edit flow
- [x] From `patient/[id]`, click "Edit Patient Details" → modal opens prefilled, DOB shows the correct date.
- [x] Same from `vitals/[id]`.
- [x] Edit name/contact/gender/allergy/DOB/cards → Save → toast success, values persist after refetch.

Manual — photo recapture
- [ ] Open edit, don't touch the photo, Save → existing photo unchanged in Cloudinary.
- [ ] Retake photo → Save → new image shows on the profile.
- [ ] After recapture, verify the patient is still found by face search (verify faceprint regenerated).

### DIFF
--- a/src/__tests__/lib/utils/patient.test.ts
+++ b/src/__tests__/lib/utils/patient.test.ts
@@ -1,4 +1,8 @@
-import { formatPatientCode, formatPatientId } from "@/lib/utils/patient";
+import {
+  formatPatientCode,
+  formatPatientId,
+  toDateInputValue,
+} from "@/lib/utils/patient";
 
 describe("formatPatientId", () => {
   it("zero-pads ids to 4 digits", () => {
@@ -14,5 +18,24 @@ describe("formatPatientId", () => {
 describe("formatPatientCode", () => {
   it("prefixes the village code when present", () => {
     expect(formatPatientCode("PC", 12)).toBe("PC0012");
+  });
+});
+
+describe("toDateInputValue", () => {
+  it("returns yyyy-MM-dd for an ISO string with a time component", () => {
+    // The shape <input type='date'> requires; the time must be dropped.
+    expect(toDateInputValue("1987-04-10T00:00:00.000Z")).toBe("1987-04-10");
+  });
+
+  it("returns yyyy-MM-dd for a bare date string", () => {
+    expect(toDateInputValue("1987-04-10")).toBe("1987-04-10");
+  });
+
+  it("returns an empty string for an unparseable date string", () => {
+    expect(toDateInputValue("not a date")).toBe("");
+  });
+
+  it("returns an empty string for an invalid Date object", () => {
+    expect(toDateInputValue(new Date(NaN))).toBe("");
   });
 });
--- a/src/__tests__/lib/utils/patient.test.ts
+++ b/src/__tests__/lib/utils/patient.test.ts
@@ -31,8 +31,30 @@ describe("toDateInputValue", () => {
     expect(toDateInputValue("1987-04-10")).toBe("1987-04-10");
   });
 
+  it("preserves single-digit month padding", () => {
+    expect(toDateInputValue("2000-01-05")).toBe("2000-01-05");
+  });
+
+  it("preserves single-digit day padding", () => {
+    expect(toDateInputValue("2000-12-01")).toBe("2000-12-01");
+  });
+
+  it("handles end-of-month dates correctly", () => {
+    expect(toDateInputValue("2024-02-29T00:00:00.000Z")).toBe("2024-02-29");
+  });
+
+  it("handles a Date object", () => {
+    expect(toDateInputValue(new Date("2005-06-15T00:00:00.000Z"))).toBe(
+      "2005-06-15",
+    );
+  });
+
   it("returns an empty string for an unparseable date string", () => {
     expect(toDateInputValue("not a date")).toBe("");
+  });
+
+  it("returns an empty string for an invalid month/day string", () => {
+    expect(toDateInputValue("2000-13-01")).toBe("");
   });
 
   it("returns an empty string for an invalid Date object", () => {
--- a/src/__tests__/server/utils/cloudinary.test.ts
+++ b/src/__tests__/server/utils/cloudinary.test.ts
@@ -1,0 +1,17 @@
+import { extractCloudinaryPublicId } from "@/server/utils/cloudinary";
+
+describe("extractCloudinaryPublicId", () => {
+  it("strips the image/upload/v{version}/ prefix", () => {
+    expect(
+      extractCloudinaryPublicId("image/upload/v1723200000/abc123xyz"),
+    ).toBe("abc123xyz");
+  });
+
+  it("handles any version number length", () => {
+    expect(extractCloudinaryPublicId("image/upload/v42/xyz")).toBe("xyz");
+  });
+
+  it("returns the input unchanged when the prefix is absent", () => {
+    expect(extractCloudinaryPublicId("abc123xyz")).toBe("abc123xyz");
+  });
+});
--- a/src/__tests__/server/utils/cloudinary.test.ts
+++ b/src/__tests__/server/utils/cloudinary.test.ts
@@ -11,7 +11,23 @@ describe("extractCloudinaryPublicId", () => {
     expect(extractCloudinaryPublicId("image/upload/v42/xyz")).toBe("xyz");
   });
 
+  it("preserves folder segments within the public id", () => {
+    expect(
+      extractCloudinaryPublicId("image/upload/v1723200000/patients/photo123"),
+    ).toBe("patients/photo123");
+  });
+
+  it("preserves a file extension in the public id", () => {
+    expect(
+      extractCloudinaryPublicId("image/upload/v1723200000/abc123.jpg"),
+    ).toBe("abc123.jpg");
+  });
+
   it("returns the input unchanged when the prefix is absent", () => {
     expect(extractCloudinaryPublicId("abc123xyz")).toBe("abc123xyz");
+  });
+
+  it("returns an empty string when given an empty string", () => {
+    expect(extractCloudinaryPublicId("")).toBe("");
   });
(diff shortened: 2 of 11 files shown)


src/components/interactive/Modal.tsx
```diff
@@ -33,17 +52,31 @@ export default function Modal({ children, onClose = () => {} }: ModalProps) {
...
+          </h2>
+          <div className="flex flex-col items-end shrink-0">
+            <div className="w-28 flex items-end justify-end">
+              <Button title="Close" colour="red" onClick={() => onClose()} />
+            </div>
+            <span className="mt-1 text-xs text-slate-400">or press Esc</span>
+          </div>
+        </div>
+      ) : (
+        <button
```
keiran0: I believe that it would be ideal to enforce that all modals contain titles (so that the modal header and behaviour is consistent across the app), but I see the need for the ternary here because it would require some refactoring and that might be a bit out of scope for the PR, thus have created a new issue here:  #209

src/components/interactive/Modal.tsx
```diff
@@ -33,17 +52,31 @@ export default function Modal({ children, onClose = () => {} }: ModalProps) {
         onClose();
       }}
       onClose={onClose}
-      className="inset-0 bg-white rounded-xl shadow-xl p-6 w-full max-w-md fixed flex min-h-60 flex-col justify-between self-center justify-self-center backdrop:fixed backdrop:inset-0 backdrop:bg-black/50"
+      className={`inset-0 bg-white rounded-xl shadow-xl p-6 w-full ${maxWidthClassName} fixed flex min-h-60 flex-col justify-between self-center justify-self-center backdrop:fixed backdrop:inset-0 backdrop:bg-black/50`}
```
keiran0: Should `clsx` be used here instead?

src/components/patient/EditPatientForm.tsx
```diff
@@ -0,0 +1,171 @@
...
+import { Button } from "@/components/interactive/Button/Button";
+import { toDateInputValue } from "@/lib/utils/patient";
+
+type Patient = NonNullable<RouterOutput["patientsRouter"]["getById"]>;
+
+type EditPatientFormValues = {
+  name: string;
+  identificationNumber: string;
+  contactNo: string;
+  gender: "male" | "female";
```
keiran0: Should this be derived from the `pgEnum` in the schema files instead?

src/lib/utils/facialRecognition.ts
```diff
@@ -34,6 +36,25 @@ export async function generateFaceprint(str: string) {
   }
 }
 
+/**
+ * Removes a faceprint from the AWS Collection.
+ * @param faceId The FaceId of the faceprint to delete.
+ */
+export async function deleteFaceprint(faceId: string) {
```
keiran0: I'm thinking of a situation where somehow two patients have the same faceprint when this function is run. Though this hypothetical scenario doesn't exist in the old database, could we implement a check to see if the cloudinary url is in use before running this function? Especially since it's a delete operation.

src/__tests__/lib/utils/patient.test.ts
```diff
@@ -16,3 +20,22 @@ describe("formatPatientCode", () => {
...
   });
 });
+
+describe("toDateInputValue", () => {
+  it("returns yyyy-MM-dd for an ISO string with a time component", () => {
+    // The shape <input type='date'> requires; the time must be dropped.
+    expect(toDateInputValue("1987-04-10T00:00:00.000Z")).toBe("1987-04-10");
+  });
+
+  it("returns yyyy-MM-dd for a bare date string", () => {
```
keiran0: Should we have more inputs for each testcase here? With different months/dates/invalid month and date etc?

src/server/utils/cloudinary.ts
```diff
@@ -20,3 +20,24 @@ export async function uploadToCloudinary(image: File) {
...
 }
+
+/**
+ * Recovers the bare public id (what the destroy API expects) from the stored
+ * `image/upload/v{version}/{public_id}` path produced by {@link uploadToCloudinary}.
+ * Any folder segments in the public id are preserved; only the
+ * `image/upload/v{version}/` prefix is stripped. If the prefix isn't present
+ * (unexpected format), the input is returned unchanged.
+ */
+export function extractCloudinaryPublicId(publicIdPath: string): string {
```
keiran0: (No action needed, just a comment to acknowledge this workaround) Created an issue #210 . I believe the prefix was included in the creation process for backward compatibility with old codebase and database, though I think it might not be necessary in the future, to reconsider this format during the database migration phase and refactor if necessary.

src/__tests__/server/utils/cloudinary.test.ts
```diff
@@ -0,0 +1,17 @@
+import { extractCloudinaryPublicId } from "@/server/utils/cloudinary";
+
+describe("extractCloudinaryPublicId", () => {
+  it("strips the image/upload/v{version}/ prefix", () => {
+    expect(
+      extractCloudinaryPublicId("image/upload/v1723200000/abc123xyz"),
+    ).toBe("abc123xyz");
+  });
+
+  it("handles any version number length", () => {
```
keiran0: Should this also contain more testcases?

keiran0 (changes requested):
Some reviews first (though not quite done):

~~Regarding the flow, would it make more sense for the pages where each patient is handled separately to link to the patient information page instead? So from vitals/vision, have a button on the top right side to go to the patient specific page `patient/id`. Then the editing can be done there instead? I'm not sure if changing patient details is something that will be done frequently for this to be such a readily available button that needs to exist on every station, but viewing a patient's past visits (which can be seen from `patient/id`) might be a more useful feature when data needs to be compared.~~ Pivoted after learning that referrals tend to edit/add the patient's phone number as well. Might as well add it then!

If the above suggestion is not taken, the size of the modal for the vitals and patient page needs to be standardized as well, they're different widths when opened on the same screen size.

src/components/interactive/Modal.tsx
derrikzzz: Used `clsx` here for className composition here [2eb207b](https://github.com/NUS-Project-SaBai/project-sabai/pull/174/commits/2eb207b5e80af3c604b08d391684ab5047e89e7d)

src/lib/utils/facialRecognition.ts
derrikzzz: Add guard at the router level [0d3f256](https://github.com/NUS-Project-SaBai/project-sabai/pull/174/commits/0d3f256f254676cc0117166c0549cc16cfb695a5)

src/__tests__/lib/utils/patient.test.ts
derrikzzz: Added more tests here [6d746e0](https://github.com/NUS-Project-SaBai/project-sabai/pull/174/commits/6d746e01b9e3f15cc6d5ac23789fb021e922ae07)

src/__tests__/server/utils/cloudinary.test.ts
derrikzzz: Added here to test for extensions and empty string here [d628d0a](https://github.com/NUS-Project-SaBai/project-sabai/pull/174/commits/d628d0a3a2649408c539fc4f321555835faef4a4)

keiran0 (changes requested):
https://github.com/user-attachments/assets/059c34a8-6c9e-4995-8ee4-d4f248ac42eb

Should the modal sizes be standardized across the same platform? Other than that, looks good!

keiran0 (approved):
LGTM!

PR #175, opened 2026-08-04
Stock editing

keiran0:
## Description

This PR adds editing stock feature to `/medication-stock`, as well as edits the router procedure to order by stock id (to prevent layout shifts during any creation/update process in the page.

Clicking the 'Edit' button at the right hand corner of every Row component will change the row to editing mode, and the editable cells for location and remarks should appear. A dropdown should also appear for the stock status. 

In the editing state, the 'Edit' button will change into a 'Save' button. Attempting to save a row with no cells edited will cause an error toast to pop up on the screen. 

A small change made to `RHFDropdown` component: adding of flex column to `RHFDropdownMenu` to prevent state dropdown from scrolling to the side.

## To be done in future PRs:

- While in the design document, the stock status is said to be not selectable by users since it's an internal state, it is still shown here and will be bulk-made non-selectable in a future PR when the reserved state is actually made to be in use.
- Shift the medication brands and active ingredients button to its final position on the screen (top right or bottom of screen?)

## Other considerations regarding the locked fields

It has been previously agreed that something like active ingredient, brand and expiry should not be made editable (the only reason why someone would want to edit it after adding is if it's a user error). The quantity is made somewhat 'editable' through the stock split feature, where every dosage is accounted for during stocktaking, reflected through the stock status states.


src/pages/medication-stock/index.tsx
```diff
@@ -184,4 +143,144 @@ function MedicationStockBasePage() {
...
+  const updateMutation = trpc.medicationStockRouter.update.useMutation({
+    onSuccess: () => {
+      toast.success("Successfully updated!");
+      utils.medicationStockRouter.listWithBrandAndActiveIngredient.invalidate();
+      form.reset();
+      setIsEditing(false);
+    },
+    onError: (err) => {
+      console.error(err);
+      toast.error("An error has occurred.");
```
derrikzzz: nit: would it be better to have a more specific error message here?

src/server/routers/medication_stock_router.ts
derrikzzz: If only `location`, `stockStatus`, and `remarks` are to be editable here, can this be trimmed further?

src/__tests__/pages/medication-stock/index.test.tsx
```diff
@@ -387,4 +394,186 @@ describe("MedicationStockPage", () => {
...
+    );
+    await user.click(dropdownButtonDonated!);
+
+    await user.click(screen.getByRole("button", { name: "Save" }));
+
+    const toastEl = screen.getByRole("status");
+    expect(toastEl).toBeInTheDocument();
+    expect(toastEl).toHaveTextContent("Successfully updated!");
+  });
 });
```
derrikzzz, line 662: Would it be better if we can have tests for 2 cases:
1. That an error toast will be shown when the update mutation fails here
2. That locked fields prob won't be part of the update payload (can use something like `expect(payload).not.toHaveProperty("expiry");`) as an example

src/server/routers/medication_stock_router.ts
keiran0: Good catch, fixed 35d87db

src/pages/medication-stock/index.tsx
keiran0: Resolved in b87f4b0

src/__tests__/pages/medication-stock/index.test.tsx
keiran0: Added in fa8ff17 and 9f84202

src/pages/medication-stock/index.tsx
```diff
@@ -184,4 +143,154 @@ function MedicationStockBasePage() {
...
+              />
+            ) : (
+              <Button
+                onClick={() => setIsEditing(true)}
+                colour="emerald"
+                title="Edit"
+              />
+            )}
+            {isEditing ? (
+              <Button
```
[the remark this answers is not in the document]
keiran0: Bug is due to the form not getting reset on editing mode exit, good catch and fixed in a66af20, test added in 33eba2b

src/pages/medication-stock/index.tsx
```diff
@@ -179,4 +138,157 @@ function MedicationStockBasePage() {
...
+        </TableCell>
+        <TableCell>
+          <span className="text-sm font-medium text-slate-900 flex flex-row gap-2">
+            {isEditing ? (
+              <Button
+                onClick={() => {
+                  form.handleSubmit(handleSubmit)();
+                }}
+                colour="emerald"
+                title="Save"
```
derrikzzz: Would it be better if we can have a `loading={updateMutation.isPending}` here, on disabling the button while the update mutation is in flight?

https://github.com/user-attachments/assets/bdacde91-42ae-48ed-8792-f2986c53ca10

keiran0: Added in cd61d37

derrikzzz (approved):
Looks good to merge in! Think it would be nice if we can have the medical students to be familiar with the actions like stock editing, stock splitting etc... haha

PR #177, opened 2026-08-04
144 stock splitting

### LINKED ISSUE #144: Medication Split Stock Feature
## Prerequisites:

- PR #130 merged to into dev

## Description

Partial implementation of the split stock feature was completed in PR #130, where splits can be added and removed from the frontend. 
The method for the splits have not been implemented yet and the current confirm button only logs the splits to the console.

## Todo

- Implement a router procedure that writes to the `medication-stock` table, deducting from the parent stock and adding new child stock with what the user has specified in the form.
- Make the modal look nicer (the `-` button for removing splits, spacing between splits etc)
- Add testcases

## Considerations

- Child stocks should maintain the same active ingredient, brand and expiry date.
- The quantity of all child stocks should add up to the parent stock. Submission should be blocked otherwise.
- The mutations should be done in the same db transaction so that the entire statement either fails or succeeds (and not just 1 split being added but the others being missing, for example)
- Splits should be distinct from each other. For example, having 2 child stocks where their locations and states are the same is basically the same as simply just editing the parent stock.

keiran0:
Closes #144

## Description

This PR creates the splitting feature for medication stock. Upon clicking the 'Split' button on a medication stock row, a modal should display where the user is able to add child splits.

The child splits will be created in the database with the respective altered states/locations/remarks/quantities specified, then the respective entries for each of these modifications will be made in the `stockChanges` table.

The original parent stock will have a quantity of 0, and will be filtered out by the `list` procedures in the router by default. This is only useful for the audit trail.

## Expected behaviour

- Clicking the 'Add Split' button should create a new split on the frontend, with default values being the parent stock's values.
- Clicking the '-' button at the side of each child split will result in that child split being removed from the screen.
- There should be at least 2 splits. Rationale is that having 1 split is the same as editing the stock (instead of splitting) and having 0 splits is the same as doing nothing.
- The total quantity of the child stocks should equal the quantity of the parent stock. If the 'Confirm' button is clicked when the quantity doesn't tally, an error toast will show up
- Clicking the 'Confirm' button when there are at least two child stocks that are identical to each other will result in an error toast being shown.

## Future considerations (Issues will be written for them when the PR is merged)

~~The current implementation assumes that the parent stock has not been modified at the time of splitting. This is a reasonable assumption as splitting would likely be done during the time of stocktaking instead of during normal operations. To consider other forms of concurrency control.~~

The second consideration is perhaps putting two child stocks by default and locking the removal of any of them so the user will automatically know that 2 stocks is a requirement, but that can be left to another future PR (enhancement). The improvement of the stock splitting UI is another thing that can also be considered.

~~The third consideration is the implementation of `areStocksDistinct`. This would be under technical debt due to the use of JSON stringify to check for uniqueness of child stocks when it's fragile to future changes (though unlikely and written as comment). To make new issue to address that.~~ Rectified in this current PR

Finally, the current implementation updates the `medicationStock` table but leaves out the changes to the `stockChanges` table as that will be handled by a database trigger and in another PR, but this is a dependency.


src/components/medication-stock/SplittingModal.tsx
```diff
@@ -61,58 +98,89 @@ export default function SplittingModal({
...
         onClick={() => setSplits([...splits, stock])}
         className="my-4"
       />
-      <Button
-        title="Confirm"
-        colour="emerald"
-        onClick={() => console.log(splits)}
-        // TODO(separate PR): validate parent qty === sum(all child qty), then bulk mutate the splits
-      />
+      <Button title="Confirm" colour="emerald" onClick={handleSubmit} />
```
derrikzzz: Would it be better if we can disable the button here while the mutation is pending?

keiran0: Fixed in 70e4675

src/lib/utils/medication-stock.ts
```diff
@@ -20,3 +26,67 @@ export type MedicationStockWithBrandAndActiveIngredient = MedicationStock & {
   medicationBrandName: string;
   medicationActiveIngredientName: string;
 };
+
+// Functions used for split stock
+
+export const splitSchema = z.object({
+  quantity: zfd.numeric(z.number().int()),
```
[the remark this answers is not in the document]
keiran0: Not too defensive at all, added in 28d071b

src/components/medication-stock/SplittingModal.tsx
```diff
@@ -35,6 +52,26 @@ export default function SplittingModal({
     setSplits(newSplits);
   }
 
+  function handleSubmit() {
+    // DO NOT CHANGE THIS MAP FUNCTION, OR THE PRESENCE OF THE REMARKS COLUMN (undefined). The key order is important for the implementation of areSplitsDistinct (due to use of json.stringify)
+    // DO NOT CHANGE THE PRESENCE OF THE REMARKS COLUMN. json.stringify drops undefined columns, so 2 objects, 1 with remarks key missing and another with remarks: undefined are evaluated the same way.
```
derrikzzz: This feels a bit fragile that we are depending on the key order here. What if we can build an array of the fields that define uniqueness here?

Maybe we can have a function like:
```
 function splitKey(split: Payload) {
    return JSON.stringify([
      split.location,
      split.stockStatus,
      split.quantity,
      split.remarks ?? "",
    ]);
  }
```

So for a split like `{ location: "Shelf A", stockStatus: "GOOD", quantity: 5, remarks: undefined }`, you will get this when you call splitKey on it:

```
["Shelf A", "GOOD", 5, ""]

(remarks ?? "" means: if remarks is undefined or null, use "" instead.)
```

Then we use `JSON.stringify` that array into a string, then let that string be the fingerprint. Now we can compare 2 splits by the keys
```
splitKey({ location: "Shelf A", stockStatus: "GOOD", quantity: 5, remarks: undefined })
// '["Shelf A","GOOD",5,""]'

splitKey({ quantity: 5, remarks: "", stockStatus: "GOOD", location: "Shelf A" })
// '["Shelf A","GOOD",5,""]'  ← identical, even though object keys  in different order
```

This will prevent issues such as accidental changes to map function here.

src/lib/utils/medication-stock.ts
```diff
@@ -20,3 +20,60 @@ export type MedicationStockWithBrandAndActiveIngredient = MedicationStock & {
...
+    const stringified = JSON.stringify(splits[i]);
+    if (splitsSet.has(stringified)) {
+      return false;
+    }
+    splitsSet.add(stringified);
+  }
+  return true;
+}
+
+export function validateSplits(
```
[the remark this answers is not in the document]
derrikzzz: Oh maybe this might be a preference thing in readability whoops, am resolving this!

src/server/routers/medication_stock_router.ts
```diff
@@ -107,4 +112,48 @@ export const medicationStockRouter = router({
...
+  createSplits: protectedProcedure
+    .input(
+      z.object({
+        parentId: zfd.numeric(z.number().int()),
+        splits: z.array(splitSchema).min(2),
+      }),
+    )
+    .mutation(async ({ input }) => {
+      const { splits, parentId } = input;
+      const parent = await db
```
derrikzzz: Currently the parent row is fetched before transaction opened, this may create a "TOCTOU", with 2 concurrent split requests on this same parent? Will acquiring a row-level lock on the parent be a better approach

src/lib/utils/medication-stock.ts
```diff
@@ -0,0 +1,71 @@
...
+      message: "Child stock quantity does not equal parent stock quantity!",
+    };
+  }
+
+  if (!areSplitsDistinct(splits)) {
+    return {
+      success: false,
+      message: "Splits are not distinct!",
+    };
+  }
```
derrikzzz: The first call to `areSplitsDistinct` on `medication-stock:26` checks this already, I believe this can be removed?

src/components/medication-stock/SplittingModal.tsx
```diff
@@ -22,6 +25,20 @@ export default function SplittingModal({
...
+
+  const splitMutation = trpc.medicationStockRouter.createSplits.useMutation({
+    onSuccess: () => {
+      toast.success("Successfully split!");
+      utils.medicationStockRouter.listWithBrandAndActiveIngredient.invalidate();
+      onClose();
+    },
+    onError: (err) => {
+      console.error(err);
+      toast.error("An error has occurred.");
```
derrikzzz: Can we be more specific about the error message here? on maybe if the server returns a meaningful error message we can use that, something like 
```
toast.error(err.message ?? "An error occurred while splitting stock.");
```

src/__tests__/pages/medication-stock/index.test.tsx
```diff
@@ -696,4 +700,343 @@ describe("MedicationStockPage", () => {
...
+        isLoading: false,
+      },
+    );
+
+    mockTrpc.medicationStockRouter.createSplits.useMutation.mockReturnValue({
+      mutate: vi.fn(),
+      isPending: false,
+    });
+
+    render(<MedicationStockBasePage />);
```
derrikzzz: This seems to be written quite a few times in this file, would it be better if we can wrap we wrap it with a `beforeEach` that will set up the `mutateMock`, `user`, the `createSplits.mutationMock` and `render` of the `<MedicationStockBasePage />`, then the tests can cover the interaction and assertions only?

src/lib/utils/medication-stock.ts
keiran0: I've removed the first call here since the rest of the conditions are faster to check for, would make sense for the function to be called at the very end since it's more expensive bdb11ef

src/components/medication-stock/SplittingModal.tsx
keiran0: Added in 36c5a90

src/__tests__/pages/medication-stock/index.test.tsx
keiran0: `user` cannot be placed into the `beforeEach` block because the subsequent references to `user` would be in the wrong scope, but I have refactored some of the duplicated components using methods similar to `ConsultForm.test.tsx` from commits 819a3dc onwards

src/server/routers/medication_stock_router.ts
keiran0: Added in 25f3b8c

src/components/medication-stock/SplittingModal.tsx
```diff
@@ -22,6 +25,22 @@ export default function SplittingModal({
...
+    onSuccess: () => {
+      toast.success("Successfully split!");
+      utils.medicationStockRouter.listWithBrandAndActiveIngredient.invalidate();
+      onClose();
+    },
+    onError: (err) => {
+      console.error(err);
+      toast.error(
+        toast.error(err.message ?? "An error occurred while splitting stock."),
+      );
```
derrikzzz: This can be simplified to the below 
```
toast.error(err.message ?? "An error occurred while splitting stock."),
```

src/__tests__/pages/medication-stock/index.test.tsx
```diff
@@ -694,6 +662,286 @@ describe("MedicationStockPage", () => {
       document.querySelector('input[name="location"]'),
     ).not.toBeInTheDocument();
     expect(screen.getByText(MOCK_STOCK[0].location)).toBeInTheDocument();
-    expect(mutateMock).not.toHaveBeenCalled();
+    expect(createMockMutation).not.toHaveBeenCalled();
```
derrikzzz: Since we are testing for updates over here (and in this case where the update mutation was not called when user clicks on cancel), should this be 
```
expect(updateMockMutation).not.toHaveBeenCalled()
```

src/__tests__/pages/medication-stock/index.test.tsx
```diff
@@ -694,6 +662,286 @@ describe("MedicationStockPage", () => {
...
+
+    renderPage();
+    await user.click(await screen.findByRole("button", { name: "Split" }));
+
+    await user.click(await screen.findByRole("button", { name: "Add Split" }));
+    await user.click(await screen.findByRole("button", { name: "Add Split" }));
+
+    await user.click(screen.getByRole("button", { name: "Confirm" }));
+
+    expect(createMockMutation).not.toHaveBeenCalled();
```
derrikzzz: Should we be using `createSplitsMockMutation` here?

src/server/routers/medication_stock_router.ts
```diff
@@ -107,4 +112,66 @@ export const medicationStockRouter = router({
 
       return result ? result : null;
     }),
+
+  createSplits: protectedProcedure
+    .input(
+      z.object({
+        parentId: zfd.numeric(z.number().int()),
+        splits: z.array(splitSchema).min(2),
```
derrikzzz: Should there be a `max()` here to prevent a single request from holding the row lock while bulk inserting an unbounded number of rows, though I'm not too sure on what should be the max limit here but we can put it loosely temporarily

src/types/medication-stock.ts
```diff
@@ -20,3 +26,15 @@ export type MedicationStockWithBrandAndActiveIngredient = MedicationStock & {
   medicationBrandName: string;
   medicationActiveIngredientName: string;
 };
+
+// Functions used for split stock
```
derrikzzz: Should the comment here be changed to `// Split stock schema and types` or it can be removed since the `splitSchema` below is kind of self-documenting?

src/types/medication-stock.ts
```diff
@@ -20,3 +26,15 @@ export type MedicationStockWithBrandAndActiveIngredient = MedicationStock & {
...
+
+export const splitSchema = z.object({
+  quantity: zfd.numeric(z.number().int().positive()),
+  location: zfd.text(z.string()),
+  stockStatus: zfd.text(z.enum(medicationStatusEnum.enumValues)),
+  remarks: zfd.text(z.string().optional()),
+});
+
+export type Payload = z.infer<typeof splitSchema>;
```
derrikzzz: `Payload` seems to be too generic and may cause some trip up if a second mutation type were to be added to this file. Let's make this type name self-documenting, perhaps this can be renamed to `SplitPayload`?

src/lib/utils/medication-stock.ts
```diff
@@ -0,0 +1,64 @@
...
+        "Please use the editing function instead to edit a single split!",
+    };
+  }
+
+  const quantity = splits.reduce(
+    (accumulator, current) => accumulator + current.quantity,
+    0,
+  );
+
+  if (quantity != parentQty) {
```
derrikzzz: Should we prefer strict equality (or in this case, strict inequality) over loose inequality here?

src/__tests__/pages/medication-stock/index.test.tsx
keiran0: Oops fixed in f3295a3

src/__tests__/pages/medication-stock/index.test.tsx
keiran0: Oops again, fixed in f3295a3

src/components/medication-stock/SplittingModal.tsx
keiran0: Think I copied pasted too hard, fixed in 7597aa9

src/types/medication-stock.ts
keiran0: Fixed in 35e3a3c

src/types/medication-stock.ts
keiran0: 57eb840 removed following the `Payload` variable rename

src/lib/utils/medication-stock.ts
keiran0: Fixed in f936396

src/server/routers/medication_stock_router.ts
keiran0: Added `max()` and corresponding block in `validateSplits` in 57d0970

src/components/medication-stock/SplittingModal.tsx
```diff
@@ -110,8 +181,8 @@ export default function SplittingModal({
       <Button
         title="Confirm"
         colour="emerald"
-        onClick={() => console.log(splits)}
-        // TODO(separate PR): validate parent qty === sum(all child qty), then bulk mutate the splits
+        onClick={handleSubmit}
+        disabled={splitMutation.isPending}
       />
```
derrikzzz: nit: I don't think it needs to be covered in this PR but would be good if the "Add Split" button can be disabled if user were to add more than 10 splits, would you be able to help create a issue for this? Rationale is that it will let the user know earlier compared to only knowing the limit after clicking on the "Confirm" button.

Image

derrikzzz: Have created an issue for this here https://github.com/NUS-Project-SaBai/project-sabai/issues/219, will resolve this first!

src/__tests__/pages/medication-stock/index.test.tsx
```diff
@@ -694,6 +662,286 @@ describe("MedicationStockPage", () => {
...
+    await user.type(quantityInputs[0], String(halfQuantity));
+    await user.clear(quantityInputs[1]);
+    await user.type(
+      quantityInputs[1],
+      String(MOCK_STOCK[0].quantity - halfQuantity),
+    );
+
+    await user.click(screen.getByRole("button", { name: "Confirm" }));
+
+    expect(createMockMutation).not.toHaveBeenCalled();
```
derrikzzz: This should be `createSplitsMockMutation` instead of `createMockMutation`?

src/lib/utils/medication-stock.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (seenKeys.has(key)) {
+      return false;
+    }
+    seenKeys.add(key);
+  }
+
+  return true;
+}
+
+export function validateSplits(splits: SplitPayload[], parentQty: number) {
```
derrikzzz: Had this thought, since now this is a new helper function that we are introducing the validate splits, should we add unit tests for this?

src/server/routers/medication_stock_router.ts
```diff
@@ -107,4 +112,66 @@ export const medicationStockRouter = router({
 
       return result ? result : null;
     }),
+
+  createSplits: protectedProcedure
+    .input(
+      z.object({
+        parentId: zfd.numeric(z.number().int()),
+        splits: z.array(splitSchema).min(2).max(10),
```
derrikzzz: Would it be better if we have `src/lib/constants/medication-stock.ts` with `MIN_SPLITS = 2` and `MAX_SPLITS = 10` defined itself, rationale is that these constants can be used at the router level here instead of magic numbers and we can also use these constants in the `validateSplits` method in `src/lib/utils/medication-stock.ts`

src/__tests__/pages/medication-stock/index.test.tsx
keiran0: 069316b oops

src/server/routers/medication_stock_router.ts
keiran0: Good call, d5eb367

src/lib/utils/medication-stock.ts
keiran0: Added in d307b97

src/lib/utils/medication-stock.ts
```diff
@@ -0,0 +1,72 @@
...
+  const seenKeys = new Set<string>();
+  for (const split of splits) {
+    const key = splitKey(split);
+    if (seenKeys.has(key)) {
+      return false;
+    }
+    seenKeys.add(key);
+  }
+
+  return true;
```
derrikzzz: nit: would it be better if we can make this more concise? We can return a new set here where we are returning `new Set(splits.map(splitKey)).size == splits.length`

src/lib/utils/medication-stock.ts
```diff
@@ -0,0 +1,72 @@
...
+      success: false,
+      message:
+        "Please use the editing function instead to edit a single split!",
+    };
+  }
+
+  if (splits.length > MAX_SPLITS) {
+    return {
+      success: false,
+      message: "Please do not exceed " + MAX_SPLITS + " splits!",
```
derrikzzz: Would it be better if we use template literals here instead of string concat?

src/__tests__/pages/medication-stock/index.test.tsx
```diff
@@ -61,20 +65,43 @@ vi.mock("@/utils/trpc", () => ({
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const mockTrpc = trpc as any;
 
+let createMockMutation: ReturnType<typeof vi.fn>;
```
derrikzzz: Is `createMockMutation` dead here?

src/__tests__/pages/medication-stock/index.test.tsx
```diff
@@ -694,6 +660,286 @@ describe("MedicationStockPage", () => {
...
+        },
+        {
+          location: "Location B",
+          stockStatus: MOCK_STOCK[0].stockStatus,
+          quantity: remainingQty,
+          remarks: MOCK_STOCK[0].remarks ?? undefined,
+        },
+      ],
+    });
   });
```
derrikzzz: Should we have tests here to assert closing the modal when the split mutation is success, and what it will be showing when the split mutation fails (an error toast)?

src/lib/utils/medication-stock.ts
derrikzzz: Template literals used here [e4b05fc](https://github.com/NUS-Project-SaBai/project-sabai/pull/177/commits/e4b05fc2cc274f76c674a35c8931479ffb112104)

src/__tests__/pages/medication-stock/index.test.tsx
derrikzzz: Removed `createMockMutation` here [0554342](https://github.com/NUS-Project-SaBai/project-sabai/pull/177/commits/0554342165ab8318285425b55207ee36ef14c9cf)

src/lib/utils/medication-stock.ts
keiran0: Fixed in e4b05fc

src/lib/utils/medication-stock.ts
keiran0: Added in 40a9059

src/__tests__/pages/medication-stock/index.test.tsx
keiran0: Good catch, must have forgotten to remove it after the renaming stuff. Fixed in 0554342

src/__tests__/pages/medication-stock/index.test.tsx
keiran0: Fixed in 2732d14 and e23b7e2

derrikzzz (approved):
Looks good to merge in, thanks for working on the stock splitting!

Kairos-T:
im interested in stocks

PR #180, opened 2026-08-05
Remarks enum

keiran0:
## Description

Previously, `remarks` was added to the `medicationStock` table. This is an editable field that should have been added to the `stockChangeFieldEnum`.


derrikzzz (approved):
Looks good to merge in! Good catch

PR #185, opened 2026-08-09
184 allow decimal values in vitals form numeric fields

derrikzzz:
Closes #184 

## Description
- Fixes bug in the patient vitals form where numeric fields rejected decimal values. Because the inputs rendered as `<input type="number">` with the default `step="1"`, entering a body temperature of `37.5 °C` failed with the browser error *"Please enter a valid value. The two nearest valid values are 37 and 38."* This forced staff to round measurements that clinically require decimals. Ideally there should be no rounding for such fields

## Changes

Frontend — `src/pages/vitals/[id]/index.tsx`
- Added a per-field `step` to the numeric inputs so decimals are accepted at the correct precision:
  - `0.1` — Height (cm), Body Temperature (°C)
  - `0.01` — Weight (kg), Fasting/Non-Fasting Blood Glucose (mmol/L), HbA1c (%), Hemocue Count (g/dL)
  - Systolic, Diastolic, and Heart Rate are remain whole numbers.

Backend — `src/server/routers/vitals_router.ts`
- Added `numericColumn` / `integerColumn` helpers that validate each field's type and precision server-side (precision mirrors the form `step`), then serialize numeric values back to the string Drizzle `numeric` columns expect.
- Error messages name the offending field

Screenshot 2026-08-09 at 2 59 20 PM

### DIFF
--- a/src/server/routers/vitals_router.ts
+++ b/src/server/routers/vitals_router.ts
@@ -27,26 +27,47 @@ const selectVitalsFields = {
 };
 
 /**
+ * Validator for PostgreSQL `numeric` column. Validates that the value is a
+ * number at the given precision (`step` mirrors the form input's `step`), then
+ * serializes back to the string that Drizzle `numeric` columns expect.
+ * `label` is used in error messages
+ *
+ */
+const numericColumn = (label: string, { step }: { step: number }) =>
+  z.coerce
+    .number({ error: `${label} must be a number` })
+    .multipleOf(step, { error: `${label} must be in steps of ${step}` })
+    .transform((n) => n.toString())
+    .optional();
+
+/** Validator for an integer column. */
+const integerColumn = (label: string) =>
+  z.coerce
+    .number({ error: `${label} must be a number` })
+    .int({ error: `${label} must be a whole number` })
+    .optional();
+
+/**
  * Input validation schema for creating vitals records.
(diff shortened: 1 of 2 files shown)


src/server/routers/vitals_router.ts
```diff
@@ -26,23 +26,47 @@ const selectVitalsFields = {
   visitId: vitals.visitId,
 };
 
+/**
+ * Validator for PostgreSQL `numeric` column. Validates that the value is a
+ * number at the given precision (`step` mirrors the form input's `step`), then
+ * serializes back to the string that Drizzle `numeric` columns expect.
+ * `label` is used in error messages
+ *
```
keiran0: Small nit, there's an extra line here, should this be removed?

keiran0 (changes requested):
Just a trivial change, otherwise looks good to merge in! Thanks for spotting and fixing the bug, I have no idea how I missed testing decimal places during my initial review 🥴

src/server/routers/vitals_router.ts
derrikzzz: Have removed the extra line here [0cde10c](https://github.com/NUS-Project-SaBai/project-sabai/pull/185/commits/0cde10c407b7ea50e9dec6bf188212ad898fc35d)

keiran0 (approved):
LGTM!

PR #187, opened 2026-08-09
186  allow clearing of vitals form fields 

derrikzzz:
Closes #186 

## Description 
- Fixes a bug where a previously-saved vital measurement could not be cleared. On a visit that already had vitals, blanking a field and saving left the old value in the db. A wrong reading could never be removed, only overwritten.

## Changes
- **`src/pages/vitals/[id]/index.tsx`** — `onSubmit` now sends `null` (not `undefined`) for cleared fields, so they are explicitly written as `NULL`. 
- **`src/server/routers/vitals_router.ts`** — vitals input fields changed from `.optional()` to `.nullish()` so `null` passes validation and reaches Drizzle.
- **`src/__tests__/pages/vitals/[id]/index.test.tsx`** — add test on load existing vitals, clear the temperature field, submit, and assert the update payload contains `temperature: null`.

## Test plan

- [x] Existing vitals test suite passes.
- [ ] Manually: open a visit with saved vitals, clear a field, save, reload => field is empty.

### DIFF
--- a/src/__tests__/pages/vitals/[id]/index.test.tsx
+++ b/src/__tests__/pages/vitals/[id]/index.test.tsx
@@ -1,7 +1,7 @@
 import PatientVitalsPage from "@/pages/vitals/[id]";
 import { trpc } from "@/utils/trpc";
 import { Toaster, toast } from "react-hot-toast";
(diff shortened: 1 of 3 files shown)


keiran0 (approved):
LGTM!

PR #199, opened 2026-08-11
198 jekyll gh actions

### LINKED ISSUE #198: Add GH pages deployment for Jekyll

keiran0:
Closes #198 

## Description

This PR sets up the Jekyll config file to use remote theme (GH pages doesn't support `minimal-mistakes-jekyll` by default) and sets up the Github page url and base url. It also edits a markdown file to use relative urls (otherwise it won't work due to wrong link)

Currently, the deployment is on the branch where this PR exists, but it can be configured to use dev branch later on after merging.

## Todo after merging

Switch deploy to dev branch


docs/_config.yml
derrikzzz: Should the config key here be changed to `since: "2013"`?

docs/_config.yml
```diff
@@ -10,8 +10,8 @@
 # Review documentation to determine if you should use `theme` or `remote_theme`
 # https://mmistakes.github.io/minimal-mistakes/docs/quick-start-guide/#installing-the-theme
 
-theme                  : "minimal-mistakes-jekyll"
-# remote_theme           : "mmistakes/minimal-mistakes"
+# theme                  : "minimal-mistakes-jekyll"
+remote_theme           : "mmistakes/minimal-mistakes"
```
derrikzzz: Should we be pinning this to a specific version? Looking at the Gemfile.lock, if we were to pin the version here it would be:
```
remote_theme: "mmistakes/minimal-mistakes@4.27.3"
```

derrikzzz: Will resolve this, let's go with doing unpinned for this and this is a docs site

docs/_config.yml
keiran0: Whoops, accidentally deleted that, fixed in 8ad3a98

derrikzzz (approved):
Looks good to merge in! This would be useful to med students to see what has been implemented and what is still WIP! 🚀

PR #202, opened 2026-08-13
Move files from utils to types

keiran0:
## Description

Moves files that should have been in `types/` to `types/`. Oops.


derrikzzz (commented):
Should there be a `eyesight.ts` in types folder that houses `EyesightFormValues` found in `EyesightForm.tsx`?

derrikzzz (approved):
Looks good to merge in, this would be a good distinction between what should be in utils and what should be in types!

PR #213, opened 2026-08-21
212 editing stock empty values

### LINKED ISSUE #212: Editing stock does not actually save empty values
## Description

When medication stock is saved with empty values in the location/remarks column, a success toast shows up even though nothing has been saved and the old values persist. To fix the bug

## Acceptance Criteria

- [ ] Location field should not be empty on save
- [ ] Remarks field should allow for empty values. 
- [ ] The remarks field in the database should display null if the remarks field was saved with empty value.

keiran0:
Closes #212 

This PR fixes an editing bug with the medication stock page

### DIFF
--- a/src/server/routers/medication_stock_router.ts
+++ b/src/server/routers/medication_stock_router.ts
@@ -94,7 +94,11 @@ export const medicationStockRouter = router({
         stockStatus: zfd.text(
           z.enum(medicationStatusEnum.enumValues).optional(),
         ),
-        remarks: zfd.text(z.string().optional()),
+        remarks: z.preprocess(
+          // To insert null instead of empty string into the db
+          (val) => (val === "" || val === undefined ? null : val),
+          z.string().nullable(),
+        ),
       }),
     )
     .mutation(async ({ input }) => {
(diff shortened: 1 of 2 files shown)


src/server/routers/medication_stock_router.ts
```diff
@@ -94,7 +94,11 @@ export const medicationStockRouter = router({
         stockStatus: zfd.text(
           z.enum(medicationStatusEnum.enumValues).optional(),
         ),
-        remarks: zfd.text(z.string().optional()),
+        remarks: z.preprocess(
+          // To insert null instead of empty string into the db
+          (val) => (val === "" || val === undefined ? null : val),
+          z.string().nullable(),
+        ),
```
derrikzzz: Would it be better if we can do some form of trimming here using `.trim()`, currently I would think if I key it multiple whitespaces, it won't persist to null in the db?

keiran0: Fixed in a0eacbf

derrikzzz (approved):
Looks good to merge! Good catch on this bug 👍

